PR #155, opened 2014-12-24
Per topic partitioning bug fix and additional extension points

kurtharriger:
* Fixed bug in partitioners by topic
* Added pluggable offset writers - useful for writing plain text offsets
* Added record partition extension point - useful for updating hive metastore
* Update context when new mapper starts 
  - hadoop reuses committer and will call setupTask with new context when starting new mapper.  Verified context is being changed through the lifecycle of committer. 
  - addCounts uses this possibly old context.  Not certain the behavior is incorrect, but it could be since the output path changes when mapper changes.


amghost:
Haven't this PR been merged yet?

HenryCaiHaiying:
Not yest, there are still pending comments unresolved.

kurtharriger:
I submitted this just before vacation, then forgot to follow up.  Ill try to resolve the comments later this week.

kurtharriger:
Updated source per comments.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
```diff
@@ -560,8 +561,14 @@ public int run(String[] args) throws Exception {
 
     props.putAll(cmd.getOptionProperties("D"));
 
-    run();
-    return 0;
+    try {
+      run();
+      log.info("Exiting successful Camus run.");
+      return 0;
+    } catch(Exception ex) {
```
HenryCaiHaiying: I forgot what was the reason that you have to catch this exception here (instead of letting this exception throw popped up through the stack?)

kurtharriger:
In our particular use case each hook is largely independent and we are more concerned about getting data into s3 then if a hook fails.  Our jenkins job greps output for errors and still fails job if errors reported so we can rerun if desired, but if updating hive partition fails thats easy to fix without rerunning the entire job.

PR #158, opened 2015-01-05
Sweeper should not depend on the order of directory entries

erikvanoosten:
Currently, when sweeper analyses the source directory tree, it assumes that the first entry inside the topic directory is 1) a directory, 2) the source sub-directory.

Both are not necessarily true.
Assumption 2) fails when destination sub-directory (e.g. 'daily') lexically comes before the source sub-directory (e.g. 'recent'). (This can only happen when source and destination trees are the same.)
Assumption 1) fails under the edge case where additional files are kept in the topic directory (or in any sub-directory). I am storing some topic meta data in the topic directory so this assumption failed for me as well.

This change will make sweeper analyze the entire topic directory and not just the first entry.


HenryCaiHaiying:
Looks fine, thanks.

PR #161, opened 2015-01-12
Adding ISO-8601 parsing to JsonStringMessageDecoder

stephenarmstrong:
Allow another special case `camus.message.timestamp.format` value to pipe all timestamps to `org.joda.time.DateTime` constructor (which properly parses [ISO-8601][ISO-8601])

[ISO-8601]: http://www.iso.org/iso/home/standards/iso8601.htm

### DIFF
--- a/camus-kafka-coders/src/main/java/com/linkedin/camus/etl/kafka/coders/JsonStringMessageDecoder.java
+++ b/camus-kafka-coders/src/main/java/com/linkedin/camus/etl/kafka/coders/JsonStringMessageDecoder.java
@@ -5,6 +5,7 @@ import com.google.gson.JsonParser;
 import com.linkedin.camus.coders.CamusWrapper;
 import com.linkedin.camus.coders.MessageDecoder;
 import org.apache.log4j.Logger;
+import org.joda.time.DateTime;
 import org.joda.time.format.DateTimeFormatter;
 import org.joda.time.format.ISODateTimeFormat;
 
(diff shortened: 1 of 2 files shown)


camus-kafka-coders/src/main/java/com/linkedin/camus/etl/kafka/coders/JsonStringMessageDecoder.java
```diff
@@ -87,6 +88,15 @@ public void init(Properties props, String topicName) {
       else if (timestampFormat.equals("unix_milliseconds")) {
         timestamp = jsonObject.get(timestampField).getAsLong();
       }
+      // Else if timestampFormat is 'ISO-8601', parse that
+      else if (timestampFormat.equals("ISO-8601")) {
+        String timestampString = jsonObject.get(timestampField).getAsString();
+        try {
```
HenryCaiHaiying: Why the error is just logged but not thrown back to stop the processing?

stephenarmstrong: I was just being consistent with what it does when normal parsing fails. If we want different behaviour, we should probalby change all of them as a seperate pull request.

HenryCaiHaiying: I see.  I will merge your pull request.

PR #163, opened 2015-01-16
Minor clean up of the README's intro.

FelixGV:
I think this change may help slightly to reduce our exposure to spam crawlers. There isn't any benefit to exposing the actual email address versus giving the link to subscribe to the group, since we'd like people to register on the group before posting anyway.

The previous text was also slightly messy.


HenryCaiHaiying:
Thanks Felix.

PR #167, opened 2015-01-27
reset pKey when message#key is null

jimjh:
The pKey object is reused across messages, but is not being reset between each read. #getNext should empty pKey when it receives a message without a partition key.

This is kinda tricky to reproduce in a test :( Any ideas?

### DIFF
--- a/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
+++ b/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
@@ -117,6 +117,7 @@ public class KafkaReader {
         buf.get(bytes, buf.position(), origSize);
         pKey.set(bytes, 0, origSize);
       } else {
+        log.warn("Received message with null message.key(): " + msgAndOffset);
         pKey.setSize(0);
       }
 
--- a/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
+++ b/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
@@ -116,6 +116,8 @@ public class KafkaReader {
         bytes = new byte[origSize];
         buf.get(bytes, buf.position(), origSize);
         pKey.set(bytes, 0, origSize);
+      } else {
+        pKey.setSize(0);
       }
 
       key.clear();


camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
```diff
@@ -116,6 +116,8 @@ public boolean getNext(EtlKey key, BytesWritable payload, BytesWritable pKey) th
         bytes = new byte[origSize];
         buf.get(bytes, buf.position(), origSize);
         pKey.set(bytes, 0, origSize);
```
HenryCaiHaiying: Actually when do you see message.key() is null?

jimjh: It happened to us in production.

In [Message.scala](https://github.com/apache/kafka/blob/0.8.1/core/src/main/scala/kafka/message/Message.scala#L136), one of the constructors sets `key` to `null`.

HenryCaiHaiying: Can you add a logging when the key is null so we know this situation happened.

And when you set the pKey.setSize(0), does that cause any problems in the remaining code?

jimjh: Added logging!

Reading the BytesWritable source, it appears that setting size to 0 is the canonical way to reset the data. The length of the actual byte array backing the BytesWritable is its capacity, which does not change.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
```diff
@@ -116,6 +116,9 @@ public boolean getNext(EtlKey key, BytesWritable payload, BytesWritable pKey) th
...
+      } else {
+        log.debug("Received message with null message.key().");
```
HenryCaiHaiying: This needs to be a WARN category logging.  And you can also log the 'messageAndOffset' to capture more information.

jimjh:
Done!

HenryCaiHaiying:
Thanks.

PR #169, opened 2015-02-06
CachedSchemaRegistry to get the lastest schema from its cache

killerwhile:
CachedSchemaRegistry is not using its cache to return the latest schema, it is requesting it everytime instead. This might result in expensive operations. It should ensure it reads from its cache, just like for other schema's requests.


miguno:
Example effect of this patch:  The completion time of one of our Camus runs (~15 GB) was reduced from 5 hours to only 4 minutes.

HenryCaiHaiying:
Looks good, thanks.

PR #171, opened 2015-02-12
Reverting the changes as throwing compile error.

saurzcode:
As @HenryCaiHaiying suggested making a pull request reverting the changes , as it was throwing compile time error.


HenryCaiHaiying:
Thanks.

PR #172, opened 2015-02-18
etl-2572: ignoring kafka events with missing schema id

zliu41:
1. Added the ability to limit the number of retries for a missing schema id
2. Added the ability to limit the min interval between two retries for a missing schema id.
3. Added hadoop counters 'MSG_SKIPPED.SCHEMA_NOT_FOUND' and 'MSG_SKIPPED.OTHER'.
4. Added unit test for CachedSchemaRegistry (CachedSchemaRegistryTest.java)


camus-api/pom.xml
```diff
@@ -17,6 +17,10 @@
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-client</artifactId>
     </dependency>
+    <dependency>
+				<groupId>org.mockito</groupId>
```
HenryCaiHaiying: Try to use easyMock

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -2,86 +2,153 @@
...
-			return true;
-		}
-
-		@Override
-		public String toString() {
-			return "CachedSchemaTuple [topic=" + topic + ", id=" + id + "]";
-		}
-	}
+  private static final String GETSCHEMA_BYID_MAX_RETIRES = "getschema.byid.max.retries";
+  private static final String GETSCHEMA_BYID_MAX_RETIRES_DEFAULT = String
```
HenryCaiHaiying: Should the default value to be 3?

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -2,86 +2,153 @@
...
+      .valueOf(Integer.MAX_VALUE);
+  private static final String GETSCHEMA_BYID_MIN_INTERVAL_SECONDS = "getschema.byid.min.interval.seconds";
+  private static final String GETSCHEMA_BYID_MIN_INTERVAL_SECONDS_DEFAULT = "0";
```
HenryCaiHaiying: Should the default to be 1 seconds?

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -2,86 +2,153 @@
...
+
```
HenryCaiHaiying: Maybe we don't need to change the coding style of existing camus code, otherwise there will be lots of diffs.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -2,86 +2,153 @@
...
+    this.getByIdMinIntervalSeconds = Integer.parseInt(props.getProperty(
+        GETSCHEMA_BYID_MIN_INTERVAL_SECONDS,
+        GETSCHEMA_BYID_MIN_INTERVAL_SECONDS_DEFAULT));
+  }
+
+  public CachedSchemaRegistry(SchemaRegistry<S> registry) {
+    this.registry = registry;
+    this.cachedById = new ConcurrentHashMap<CachedSchemaTuple, S>();
+    this.cachedLatest = new ConcurrentHashMap<String, SchemaDetails<S>>();
+    this.getByIdMaxRetries = Integer
```
HenryCaiHaiying: How do you parse the properties without the props passed in?

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -2,86 +2,153 @@
...
+        - failedFetchHistory.previousAttemptTime >= TimeUnit.SECONDS
+        .toNanos(getByIdMinIntervalSeconds);
+    return maxRetriesNotExceeded && minRetryIntervalSatisfied;
+  }
+
+  private S fetchFromSchemaRegistry(String topic, String id) {
+    try {
+      S schema = registry.getSchemaByID(topic, id);
+      return schema;
+    } catch (SchemaNotFoundException e) {
```
HenryCaiHaiying: I think the underlying registry/restClient will return SchemaRegistryClientException instead of SchemaNotFoundException in this case.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -2,86 +2,153 @@
...
+      addFetchToFailureHistory(id);
+      throw new SchemaNotFoundException(e);
+    }
+  }
+
+  private void addFetchToFailureHistory(String id) {
```
HenryCaiHaiying: This class doesn't look like thread-safe.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -2,86 +2,153 @@
...
-				return false;
-			return true;
-		}
-
-		@Override
-		public String toString() {
-			return "CachedSchemaTuple [topic=" + topic + ", id=" + id + "]";
-		}
-	}
+  private static final String GETSCHEMA_BYID_MAX_RETIRES = "getschema.byid.max.retries";
```
HenryCaiHaiying: property name: getSchema.byId.max.retries or get.schema.by.id.max.retries

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlRecordReader.java
```diff
@@ -116,11 +126,18 @@ public synchronized void close() throws IOException {
...
   }
 
-  private CamusWrapper getWrappedRecord(String topicName, byte[] payload) throws IOException {
+  private CamusWrapper getWrappedRecord(String topicName, byte[] payload)
+      throws IOException {
     CamusWrapper r = null;
     try {
       r = decoder.decode(payload);
+    } catch (SchemaNotFoundException e) {
+      mapperContext.getCounter(MSG_SKIPPED.SCHEMA_NOT_FOUND).increment(1);
```
HenryCaiHaiying: How do we use these counter values later?  I think one of the requirement is to fail the job when the skip percentage is above certain threshold.

pom.xml
```diff
@@ -78,7 +83,12 @@
 			<dependency>
 				<groupId>org.apache.hadoop</groupId>
 				<artifactId>hadoop-client</artifactId>
-				<version>1.0.3</version>
+				<version>2.6.0</version>
```
HenryCaiHaiying: Use 2.3.0 instead.

HenryCaiHaiying: And don't include the hadoop-client jar, include individual ones instead:

      h2Compile spec.external.mapredCore
      h2Compile spec.external.mapredClientCommon
      h2Compile spec.external.hadoopCommon

HenryCaiHaiying: Hadoop dependency should be specified in top-level pom.xml, not in sub-folder.

pom.xml
```diff
@@ -45,6 +45,11 @@
 			</dependency>
 			<dependency>
 				<groupId>com.linkedin.camus</groupId>
+				<artifactId>camus-example</artifactId>
```
HenryCaiHaiying: This dependency is very weird, why does it depends on camus-example?

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
ydailinkedin: If the props is passed in, can we just call init() in the constructor?

camus-api/pom.xml
zliu41: Fixed

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
zliu41: Fixed

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
zliu41: Fixed.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
zliu41: Won't change further.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
zliu41: Added props to constructor of CachedSchemaRegistry

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
zliu41: I did a global search and there's no match for 'SchemaRegistryClientException' anywhere (no definition or usage). Also, the SchemaRegistry interface Javadoc says it throws SchemaNotFoundException.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
zliu41: Fixed.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
zliu41: Fixed.

pom.xml
zliu41: Fixed, except that it seems I have to put hadoop dependency in camus-api/pom.xml, otherwise mvn clean package fails.

pom.xml
zliu41: Fixed. Doesn't seem to be needing this in top level pom.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlRecordReader.java
zliu41: Added method "checkIfTooManySkippedMsg(counter)" in CamusJob.

pom.xml
zliu41: Removed, seems we don't need this in top level pom.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -1,87 +1,155 @@
...
+    return maxRetriesNotExceeded && minRetryIntervalSatisfied;
+  }
+
+  private S fetchFromSchemaRegistry(String topic, String id) {
+    try {
+      S schema = registry.getSchemaByID(topic, id);
+      return schema;
+    } catch (SchemaNotFoundException e) {
+      addFetchToFailureHistory(id);
+      throw new SchemaNotFoundException(e);
```
chavdar: nit: Why not rethrow the same exception? this will add another meaningless "Caused by:" section.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -1,87 +1,155 @@
...
+    boolean minRetryIntervalSatisfied =
+        System.nanoTime() - failedFetchHistory.previousAttemptTime >= TimeUnit.SECONDS
+            .toNanos(getByIdMinIntervalSeconds);
+    return maxRetriesNotExceeded && minRetryIntervalSatisfied;
+  }
+
+  private S fetchFromSchemaRegistry(String topic, String id) {
+    try {
+      S schema = registry.getSchemaByID(topic, id);
+      return schema;
```
chavdar: You have to update cachedById . Previously, there was a cachedById.putIfAbsent() which seems to have been dropped.

camus-api/src/test/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistryTest.java
```diff
@@ -0,0 +1,69 @@
...
+  public void setupRegistryMock() {
+    props = new Properties();
+    registry = EasyMock.createNiceMock(SchemaRegistry.class);
+    EasyMock.expect(registry.getSchemaByID(EasyMock.anyString(), EasyMock.anyString())).andThrow(
+        new SchemaNotFoundException());
+  }
+
+  @Test
+  public void testMaxRetries() {
+    EasyMock.expectLastCall().times(20);
```
chavdar: nit: I find it a bit strange to have the expect() method in setupRetistryMock() and then have the .times(20) here. It may create unexpected behavior if someone adds another expect() call in setupRetistryMock() . I think it is better to keep the expect() call close to times() even if it means to duplicate one line of code.

chavdar:
LGTM.

zliu41:
Accidentally merged the changes in ETL2575 into this pull request. Rolled back.

camus-api/pom.xml
```diff
@@ -14,9 +14,21 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.apache.hadoop</groupId>
-      <artifactId>hadoop-client</artifactId>
-    </dependency>
+			<groupId>org.apache.hadoop</groupId>
```
HenryCaiHaiying: indentation seems weird.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -1,87 +1,156 @@
...
+    return registry.register(topic, schema);
+  }
+
+  public S getSchemaByID(String topic, String id) {
+    CachedSchemaTuple cacheKey = new CachedSchemaTuple(topic, id);
+    S schema = cachedById.get(cacheKey);
+    if (schema == null) {
+      synchronized (this) {
+        if (shouldFetchFromSchemaRegistry(id)) {
+          schema = fetchFromSchemaRegistry(topic, id);
```
HenryCaiHaiying: Does 'schema' need to put back to cachedById cache?

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -1,87 +1,156 @@
...
+        } else {
+          throw new SchemaNotFoundException();
+        }
+      }
+    }
+    return schema;
+  }
+
+  private boolean shouldFetchFromSchemaRegistry(String id) {
```
HenryCaiHaiying: Do you want to synchronize on this method?

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -1,87 +1,156 @@
...
+    boolean minRetryIntervalSatisfied =
+        System.nanoTime() - failedFetchHistory.previousAttemptTime >= TimeUnit.SECONDS
+            .toNanos(getByIdMinIntervalSeconds);
+    return maxRetriesNotExceeded && minRetryIntervalSatisfied;
+  }
+
+  private S fetchFromSchemaRegistry(String topic, String id) {
+    try {
+      S schema = registry.getSchemaByID(topic, id);
+      cachedById.putIfAbsent(new CachedSchemaTuple(topic, id), schema);
```
HenryCaiHaiying: I think it's probably better to update 'cachedById' in the caller code where it has access to the cache.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
```diff
@@ -1,87 +1,156 @@
...
+      return schema;
+    } catch (SchemaNotFoundException e) {
+      addFetchToFailureHistory(id);
+      throw e;
+    }
+  }
+
+  private void addFetchToFailureHistory(String id) {
```
HenryCaiHaiying: Should we 'synchronize' on this method?

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
```diff
@@ -392,6 +400,31 @@ public int compare(FileStatus f1, FileStatus f2) {
     }
   }
 
+  private void checkIfTooManySkippedMsg(Counter counter) {
+    if (counter.getDisplayName().equals(EtlRecordReader.MSG_SKIPPED.SCHEMA_NOT_FOUND.toString())) {
+      long maxSkippedMsgDueToSchemaNotFound =
```
HenryCaiHaiying: It might be better to parse this config parameter outside this function.

HenryCaiHaiying: This parameters is probably a percentage value (e.g. 0.1%) instead of absolute value.

camus-example/pom.xml
```diff
@@ -74,6 +70,41 @@
         </executions>
       </plugin>
     </plugins>
+    <pluginManagement>
+    	<plugins>
+    		<!--This plugin's configuration is used to store Eclipse m2e settings only. It has no influence on the Maven build itself.-->
```
HenryCaiHaiying: Indentation is not right.

camus-example/pom.xml
```diff
@@ -74,6 +70,41 @@
...
+    		<plugin>
+    			<groupId>org.eclipse.m2e</groupId>
```
HenryCaiHaiying: Why do you need 'org.eclipse.m2e' here?

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
zliu41: I added synchronized block on line 59.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
zliu41: Moved cachedById.putIfAbsent() from method "fetchFromSchemaRegistry" to here.

zliu41:
- Change config parameters for skipped msg to percentage, instead of absolute value. (default: 10%)
- Added unittest for job failure when too many msg are skipped, due to either schemaNotFound or other. 
 (testJobFailDueToSkippedMsgSchemaNotFound() and testJobFailDueToSkippedMsgOther() in CamusJobTest.java). To do the unit test, I use a flag in EtlRecordReader. When flag==true, it loads a mocked decoder, which throws SchemaNotFoundException or RuntimeException for 30% of messages.
- Fixed other small issues.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
zliu41: I have refactored the checkIfTooManySkippedMsg() method.

camus-example/pom.xml
zliu41: Removed. For some reason this is needed to import the project into eclipse.

camus-api/pom.xml
```diff
@@ -1,23 +1,35 @@
...
+	<name>Camus API.</name>
+	<packaging>jar</packaging>
 
-  <dependencies>
-    <dependency>
-      <groupId>org.apache.hadoop</groupId>
-      <artifactId>hadoop-client</artifactId>
-    </dependency>
-  </dependencies>
+	<dependencies>
```
HenryCaiHaiying: Looks like you have 'tab' instead of 'white spaces' in this file.  The indentation doesn't seem right.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
HenryCaiHaiying: I don't know whether you really need line 59: 'synchronized (this)' since you are using ConcurrentHashmap

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
```diff
@@ -392,6 +399,49 @@ public int compare(FileStatus f1, FileStatus f2) {
     }
   }
 
+  private void checkIfTooManySkippedMsg(Counters counters) {
+    double maxPercentSkippedSchemaNotFound = Double.parseDouble(props.getProperty(ETL_MAX_PERCENT_SKIPPED_SCHEMANOTFOUND,
+        ETL_MAX_PERCENT_SKIPPED_SCHEMANOTFOUND_DEFAULT));
```
HenryCaiHaiying: The default value should be '0.1%' instead of '10%'

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
```diff
@@ -392,6 +399,49 @@ public int compare(FileStatus f1, FileStatus f2) {
...
+        CounterGroup group = counters.getGroup(groupName);
+        for (Counter counter : group) {
+          if (counter.getDisplayName().equals(EtlRecordReader.KAFKA_MSG.DECODE_SUCCESSFUL.toString())) {
+            actualDecodeSuccessful = counter.getValue();
+          } else if (counter.getDisplayName().equals(EtlRecordReader.KAFKA_MSG.SKIPPED_SCHEMA_NOT_FOUND.toString())) {
+            actualSkippedSchemaNotFound = counter.getValue();
+          } else if (counter.getDisplayName().equals(EtlRecordReader.KAFKA_MSG.SKIPPED_OTHER.toString())) {
+            actualSkippedOther = counter.getValue();
+          }
+        }
```
HenryCaiHaiying: If 'actualSkipped' > 'actualDecodeSuccessful', should throw an exception.

zliu41:
- Refactored TestEtlInputFormat classes

There were two TestEtlInputFormat classes before, and when I tried to do unit test for ETL2572, I realized I either need a third TestEtlInputFormat class (because I needed to modify how a particular method in EtlInputFormat behaves), or change the constructor parameters of one of them (e.g. add a boolean to indicate behavior). Neither of them is scalable, since if we add more unit tests in future, we may end up having lots of TestEtlInputFormatClass, or lots of parameters in the constructor.

So I combined the two existing TestEtlInputFormat into a single top-level class EtlInputFormatForUnitTest. I made some redesign, see notes in that file for details.

- Added more unit tests for ETL2572

Currently have four: job failure due to too many msg skipped (schema not found / other reason); job succeed even though a small number of msg skipped (schema not found / other reason)

All unit tests passed.

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
zliu41: Because if a thread starts to execute line 61, other thread should get false from line 60. So I think we need to put 60 and 61 in one synchronized block.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
zliu41: `actualSkipped > actualDecodeSuccessful` simply means more than half of the msg were skipped. It will be caught when camus job examines the percentage of skipped messages (as long as the max allowed percentage < 50%)

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlRecordReader.java
```diff
@@ -231,7 +246,7 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
               new KafkaReader(inputFormat, context, request, CamusJob.getKafkaTimeoutValue(mapperContext),
                   CamusJob.getKafkaBufferSize(mapperContext));
 
-          decoder = MessageDecoderFactory.createMessageDecoder(context, request.getTopic());
+          decoder = createDecoder(request.getTopic());
```
zliu41: define a method `createDecoder` so that it can be overridden.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -88,7 +90,7 @@ public static void beforeClass() throws IOException {
 
     // You can't delete messages in Kafka so just writing a set of known messages that can be used for testing
     messagesWritten = new HashMap<String, List<MyMessage>>();
-    messagesWritten.put(TOPIC_1, writeKafka(TOPIC_1, 2));
+    messagesWritten.put(TOPIC_1, writeKafka(TOPIC_1, 10));
```
zliu41: Change to 10 because I want to test 30% of the message skipped (3 skipped, 7 succeeded)

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -365,20 +437,20 @@ private void assertCamusContains(String topic, List<MyMessage> messages) throws
...
-    public static void setLogger(Logger log) {
-      EtlInputFormat.setLogger(log);
-    }
-    
-    @Override
-    public SimpleConsumer createSimpleConsumer(JobContext context, String host, int port) {
-      return _simpleConsumer;
-    }
-  }
+//  private static class TestEtlInputFormat extends EtlInputFormat {
```
zliu41: no longer used, now using top level class EtlInputFormatForUnitTest

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatForUnitTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.easymock.EasyMock;
+
+import com.linkedin.camus.coders.CamusWrapper;
+import com.linkedin.camus.etl.kafka.CamusJob;
+import com.linkedin.camus.etl.kafka.common.EtlKey;
+
+
+public class EtlInputFormatForUnitTest extends EtlInputFormat {
+  public static enum ConsumerType {
```
zliu41: using enum to control behavior, e.g., whether we mock consumer, whether we load a different record reader class, etc. If future unit tests need to modify other behaviors of EtlInputFormat, simply add more such enums.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlRecordReaderForUnitTest.java
```diff
@@ -0,0 +1,59 @@
...
+import com.linkedin.camus.coders.CamusWrapper;
+import com.linkedin.camus.coders.MessageDecoder;
+import com.linkedin.camus.etl.kafka.coders.MessageDecoderFactory;
+import com.linkedin.camus.schemaregistry.SchemaNotFoundException;
+
+
+public class EtlRecordReaderForUnitTest extends EtlRecordReader {
+
+  public static enum DecoderType {
+    REGULAR,
```
zliu41: same idea as EtlInputFormatForUnitTest. use enum to indicate desired behavior (in this case, what decoder should this class use).

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -102,7 +104,6 @@ public static void afterClass() {
 
   @Before
   public void before() throws IOException, NoSuchFieldException, IllegalAccessException {
-    resetCamus();
```
zliu41: resetCamus() is removed. The reset is now done in the After method.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -136,10 +137,17 @@ public void before() throws IOException, NoSuchFieldException, IllegalAccessExce
...
+  public void after() throws IOException, SecurityException, NoSuchFieldException,
+    IllegalArgumentException, IllegalAccessException {
     // Delete all camus data
     folder.delete();
-    _mocks.clear();
+    mocks.clear();
+    EtlInputFormatForUnitTest.consumerType = EtlInputFormatForUnitTest.ConsumerType.REGULAR;
+    EtlInputFormatForUnitTest.recordReaderClass = EtlInputFormatForUnitTest.RecordReaderClass.REGULAR;
+    EtlRecordReaderForUnitTest.decoderType = EtlRecordReaderForUnitTest.DecoderType.REGULAR;
+    Field field = EtlMultiOutputFormat.class.getDeclaredField("committer");
```
zliu41: After each test, reset all behavior to REGULAR, so that settings in the current test won't affect future tests.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -207,26 +216,27 @@ private FetchResponse mockFetchResponse(List<MyMessage> myMessages) {
     }
     ByteBufferMessageSet messageSet = new ByteBufferMessageSet(messages);
     EasyMock.expect(fetchResponse.messageSet(EasyMock.anyString(), EasyMock.anyInt())).andReturn(messageSet).times(1);
-    _mocks.add(fetchResponse);
+    mocks.add(fetchResponse);
     return fetchResponse;
   }
 
-  private void mockSimpleConsumer(TopicMetadataResponse metadataResponse, OffsetResponse offsetResponse,
+  private SimpleConsumer mockSimpleConsumer(TopicMetadataResponse metadataResponse, OffsetResponse offsetResponse,
```
zliu41: The method now returns SimpleConsumer rather than void, because the return value needs to be passed to EtlInputFormatForUnitTest.consumer

camus-api/src/main/java/com/linkedin/camus/schemaregistry/CachedSchemaRegistry.java
HenryCaiHaiying: Add a javadoc note indicate that caller needs to do the synchronization.

On the other hand, it probably doesn't hurt to have 'synchronized' here since the caller already holds the same lock.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
HenryCaiHaiying: Looks good now.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
HenryCaiHaiying: OK.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatForUnitTest.java
HenryCaiHaiying: Looks good.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlRecordReaderForUnitTest.java
HenryCaiHaiying: Good

PR #175, opened 2015-02-21
etl2566 - handling offset range call failure

zliu41:
If unable to fetch latest offset from leader (response==null, response.hasError(), or caught runtime exception), retry three times. If still unable, skip the corresponding topics, proceed with other topics and report job failure.

### DIFF
--- a/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
+++ b/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
@@ -79,7 +79,7 @@ public class EtlInputFormat extends InputFormat<EtlKey, CamusWrapper> {
   public static final String CAMUS_WORK_ALLOCATOR_CLASS = "camus.work.allocator.class";
   public static final String CAMUS_WORK_ALLOCATOR_DEFAULT = "com.linkedin.camus.workallocater.BaseAllocator";
 
-  public static final int FETCH_FROM_LEADER_MAX_RETRIES = 2;
+  public static final int FETCH_FROM_LEADER_MAX_RETRIES = 3;
 
   public static boolean reportJobFailureDueToSkippedMsg = false;
 
--- a/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
+++ b/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
@@ -25,7 +25,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
+
 import kafka.api.PartitionOffsetRequestInfo;
 import kafka.common.ErrorMapping;
 import kafka.common.TopicAndPartition;
@@ -35,6 +37,7 @@ import kafka.javaapi.PartitionMetadata;
 import kafka.javaapi.TopicMetadata;
 import kafka.javaapi.TopicMetadataRequest;
 import kafka.javaapi.consumer.SimpleConsumer;
+
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
@@ -75,6 +78,10 @@ public class EtlInputFormat extends InputFormat<EtlKey, CamusWrapper> {
 
   public static final String CAMUS_WORK_ALLOCATOR_CLASS = "camus.work.allocator.class";
   public static final String CAMUS_WORK_ALLOCATOR_DEFAULT = "com.linkedin.camus.workallocater.BaseAllocator";
+
+  public static final int FETCH_FROM_LEADER_MAX_RETRIES = 2;
+
+  public static boolean reportJobFailureDueToSkippedMsg = false;
 
   private static Logger log = null;
 
@@ -176,13 +183,18 @@ public class EtlInputFormat extends InputFormat<EtlKey, CamusWrapper> {
         earliestOffsetInfo.put(topicAndPartition, partitionEarliestOffsetRequestInfo);
       }
 
-      OffsetResponse latestOffsetResponse =
-          consumer.getOffsetsBefore(new OffsetRequest(latestOffsetInfo, kafka.api.OffsetRequest.CurrentVersion(),
-              CamusJob.getKafkaClientName(context)));
-      OffsetResponse earliestOffsetResponse =
-          consumer.getOffsetsBefore(new OffsetRequest(earliestOffsetInfo, kafka.api.OffsetRequest.CurrentVersion(),
-              CamusJob.getKafkaClientName(context)));
+      OffsetResponse latestOffsetResponse = getLatestOffsetResponse(consumer, latestOffsetInfo, context);
+      OffsetResponse earliestOffsetResponse = null;
+      if (latestOffsetResponse != null) {
+        earliestOffsetResponse = getLatestOffsetResponse(consumer, earliestOffsetInfo, context);
+      }
       consumer.close();
+      if (earliestOffsetResponse == null) {
+        log.warn(generateLogWarnForSkippedTopics(earliestOffsetInfo, consumer));
+        reportJobFailureDueToSkippedMsg = true;
+        continue;
+      }
+
       for (TopicAndPartition topicAndPartition : topicAndPartitions) {
         long latestOffset = latestOffsetResponse.offsets(topicAndPartition.topic(), topicAndPartition.partition())[0];
         long earliestOffset =
@@ -198,6 +210,34 @@ public class EtlInputFormat extends InputFormat<EtlKey, CamusWrapper> {
       }
     }
     return finalRequests;
+  }
+
+  private OffsetResponse getLatestOffsetResponse(SimpleConsumer consumer,
+      Map<TopicAndPartition, PartitionOffsetRequestInfo> offsetInfo, JobContext context) {
+    for (int i = 0; i <= FETCH_FROM_LEADER_MAX_RETRIES; i++) {
+      try {
+        OffsetResponse offsetResponse = consumer.getOffsetsBefore(new OffsetRequest(offsetInfo,
+            kafka.api.OffsetRequest.CurrentVersion(), CamusJob.getKafkaClientName(context)));
+        if (offsetResponse == null || offsetResponse.hasError()) {
+          throw new RuntimeException();
+        }
+        return offsetResponse;
+      } catch (RuntimeException e) {
+        log.warn("Fetching offset from leader " + consumer.host() + ":" + consumer.port()
+            + " has failed " + (i + 1) + " time(s). " + (FETCH_FROM_LEADER_MAX_RETRIES - i) + " retries left.");
+      }
+    }
+    return null;
+  }
+
+  private String generateLogWarnForSkippedTopics(Map<TopicAndPartition, PartitionOffsetRequestInfo> offsetInfo, SimpleConsumer consumer) {
+    StringBuilder sb = new StringBuilder();
+    sb.append("The following topics will be skipped due to failure in fetching latest offsets from leader "
+        + consumer.host() + ":" + consumer.port());
+    for (TopicAndPartition topicAndPartition : offsetInfo.keySet()) {
+      sb.append("  " + topicAndPartition.topic());
+    }
+    return sb.toString();
   }
 
   public String createTopicRegEx(HashSet<String> topicsSet) {
--- a/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
+++ b/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
@@ -222,9 +222,17 @@ public class EtlInputFormat extends InputFormat<EtlKey, CamusWrapper> {
           throw new RuntimeException();
         }
         return offsetResponse;
-      } catch (RuntimeException e) {
+      } catch (Exception e) {
         log.warn("Fetching offset from leader " + consumer.host() + ":" + consumer.port()
             + " has failed " + (i + 1) + " time(s). " + (FETCH_FROM_LEADER_MAX_RETRIES - i) + " retries left.");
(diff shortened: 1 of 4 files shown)


camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
```diff
@@ -176,13 +183,18 @@ private SimpleConsumer createConsumer(JobContext context, String broker) {
...
-      OffsetResponse earliestOffsetResponse =
-          consumer.getOffsetsBefore(new OffsetRequest(earliestOffsetInfo, kafka.api.OffsetRequest.CurrentVersion(),
-              CamusJob.getKafkaClientName(context)));
+      OffsetResponse latestOffsetResponse = getLatestOffsetResponse(consumer, latestOffsetInfo, context);
+      OffsetResponse earliestOffsetResponse = null;
+      if (latestOffsetResponse != null) {
+        earliestOffsetResponse = getLatestOffsetResponse(consumer, earliestOffsetInfo, context);
+      }
       consumer.close();
+      if (earliestOffsetResponse == null) {
```
HenryCaiHaiying: Why there is no warning generation for latestOffsetResponse?

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
```diff
@@ -200,6 +212,34 @@ private SimpleConsumer createConsumer(JobContext context, String broker) {
...
+      Map<TopicAndPartition, PartitionOffsetRequestInfo> offsetInfo, JobContext context) {
+    for (int i = 0; i <= FETCH_FROM_LEADER_MAX_RETRIES; i++) {
+      try {
+        OffsetResponse offsetResponse = consumer.getOffsetsBefore(new OffsetRequest(offsetInfo,
+            kafka.api.OffsetRequest.CurrentVersion(), CamusJob.getKafkaClientName(context)));
+        if (offsetResponse == null || offsetResponse.hasError()) {
+          throw new RuntimeException();
+        }
+        return offsetResponse;
+      } catch (RuntimeException e) {
```
HenryCaiHaiying: The Kafka exception is not RuntimeException

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
```diff
@@ -200,6 +212,34 @@ private SimpleConsumer createConsumer(JobContext context, String broker) {
...
+        log.warn("Fetching offset from leader " + consumer.host() + ":" + consumer.port()
```
HenryCaiHaiying: Need to put a sleep during iterations.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
zliu41: For method 'generateLogWarnForSkippedTopics()', it doesn't matter whether we pass 'earliestOffsetInfo' or 'latestOffsetInfo' to it. Because it just needs to print the skipped topics, and 'earliestOffsetInfo' and 'latestOffsetInfo' contain exactly the same topics.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
zliu41: Fixed.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
zliu41: Fixed.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
HenryCaiHaiying: But it can still happen that earliestOffsetResponse != null but latestOffsetResponse == null since you are making two different kafka calls at different times, I think you need to check for both.

zliu41: In the 'if' block in line 188-190, it retrieves earliest only if latest!=null. So if latest=null, earliest will be null.

HenryCaiHaiying: But if latest == null, you don't have earliestOffsetInfo, then what are you going to log in line 193?

zliu41: if latest==null, I don't have earliestOffsetResponse, but I do have earliestOffsetInfo, which is set in the for loop (181-184)

HenryCaiHaiying: Looks good then.

zliu41:
Added unit test for offset range call failure:
- Use a flag in EtlInputFormat (useMockConsumerForUnitTest). If flag==true, it will use a mocked consumer, which throws exception when trying to retrieve offsets.

Also addressed other minor issues.

zliu41:
Rewrote unit test using the new mock framework. Besides the original test that throws exception for offset range calls, added two new tests:
- offsetResponse always has error.
- offsetResponse has error for the first 3 times, but succeeds the last time.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -151,7 +154,15 @@ public void runJob1() throws Exception {
   }
 
   private void setupJob1() {
```
zliu41: Broke the original setupJob1() into multiple small methods, each creating one mock, so that they can be reused.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -137,12 +139,13 @@ public void before() throws IOException, NoSuchFieldException, IllegalAccessExce
   public void after() throws IOException {
     // Delete all camus data
     folder.delete();
+    _mocks.clear();
```
zliu41: _mocks.clear() is needed because _mocks is used by multiple tests.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -193,25 +208,117 @@ private void setupJob1() {
...
     _mocks.add(fetchResponse);
-    
-    // For SimpleConsumer
-    _simpleConsumer = EasyMock.createMock(SimpleConsumer.class);
+    return fetchResponse;
+  }
+
+  private void mockSimpleConsumer(TopicMetadataResponse metadataResponse, OffsetResponse offsetResponse,
+      FetchResponse fetchResponse) {
+    _simpleConsumer = EasyMock.createNiceMock(SimpleConsumer.class);
```
zliu41: changed createMock() to createNiceMock(), because the new code needs to call consumer.host() and consumer.port() in generating log messages, which are not defined for the mocked consumer.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -193,25 +208,117 @@ private void setupJob1() {
...
     _mocks.add(_simpleConsumer);
     EasyMock.expect(_simpleConsumer.send((TopicMetadataRequest)EasyMock.anyObject())).andReturn(metadataResponse).times(1);
-    EasyMock.expect(_simpleConsumer.getOffsetsBefore((OffsetRequest)EasyMock.anyObject())).andReturn(offsetResponse).times(2);
+    EasyMock.expect(_simpleConsumer.getOffsetsBefore((OffsetRequest)EasyMock.anyObject())).andReturn(offsetResponse).anyTimes();
```
zliu41: changed times(2) to anyTimes(), because it needs to be called more than twice in the new code (max retry = 3)

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -170,17 +181,21 @@ private void setupJob1() {
...
-    // For OffsetResponse
+  private OffsetResponse mockOffsetResponse(List<MyMessage> myMessages) {
     OffsetResponse offsetResponse = EasyMock.createMock(OffsetResponse.class);
     _mocks.add(offsetResponse);
-    List<MyMessage> myMessages = messagesWritten.get(TOPIC_1);
     // The first call is getLatestOffset, we set the value to 1
     EasyMock.expect(offsetResponse.offsets(EasyMock.anyString(), EasyMock.anyInt())).andReturn(new long[]{myMessages.size()}).times(1);
     // The second call is getEarliestOffset, we set the value to 0
     EasyMock.expect(offsetResponse.offsets(EasyMock.anyString(), EasyMock.anyInt())).andReturn(new long[]{0}).times(1);
+    EasyMock.expect(offsetResponse.hasError()).andReturn(false).anyTimes();
```
zliu41: This is added because otherwise, hasError() will return true by default.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -193,25 +208,117 @@ private void setupJob1() {
...
+    OffsetResponse offsetResponse = EasyMock.createNiceMock(OffsetResponse.class);
+    _mocks.add(offsetResponse);
+    EasyMock.expect(offsetResponse.offsets(EasyMock.anyString(), EasyMock.anyInt())).andReturn(new long[]{myMessages.size()}).times(1);
+    EasyMock.expect(offsetResponse.offsets(EasyMock.anyString(), EasyMock.anyInt())).andReturn(new long[]{0}).times(1);
+    EasyMock.expect(offsetResponse.hasError()).andReturn(true).anyTimes();
+    return offsetResponse;
+  }
+
+  @Test
+  public void testJobOffsetRangeCallThirdRetrySucceed() throws Exception {
```
zliu41: this test doesn't have verify() or assert(), because it passes if the job succeeds (after three retries).

zliu41:
Regression test passed on Nertz.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
HenryCaiHaiying: Looks good.

HenryCaiHaiying: Or the _mocks shouldn't be an instance variable.  Since we are not sharing mocks between different tests, it probably should be method level variable.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
HenryCaiHaiying: Why not passing the expected return value as an argument to mockOffsetResponse(), this way you can control whether the mock object should return true or false.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
HenryCaiHaiying: I prefer the createMock because it is more strict.  Can we just add consumer.host(), consumer.port() in the mocked consumer object?

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
HenryCaiHaiying: Can the 'times' be an argument so each test can set exactly how many times it was expecting?

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
HenryCaiHaiying: Even for three retries, can we calculate how many times the mock will be called for verify() to pass?

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
zliu41: Some tests may have more complex requirements, e.g., first two times has error, third time succeeds. Also, some tests may need to change the return value of a different method, rather than hasError(). 
So, I think there should be a separate method for each need. Currently there are 3: mockOffsetResponse(), mockOffsetResponseWithError() and mockOffsetResponseThridRetrySucceed().

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
zliu41: For the same reason, I think it shouldn't be an argument. We can have different methods, e.g., mockSimpleConsumer(), mockSimpleConsumerThrowsExceptionForOffsetRangeCall(), etc.

zliu41:
- Replaced all nice mocks with regular mocks
- use anyTimes() as few as possible.
- no longer verify exception messages.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
zliu41: Still keeping _mocks as instance variable. Otherwise, a lot of methods need to add a parameter, which is cumbersome.

PR #176, opened 2015-02-21
etl2574: handling camus offset out of range AND etl2573: handling error in KafkaReader.fetch()  

zliu41:
If out of range, skip topic and report job failure.


camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
```diff
@@ -163,53 +168,91 @@ public boolean fetch() throws IOException {
...
-          MessageAndOffset skippedMessage = messageIter.next();
-          log.debug("Skipping offset : " + skippedMessage.offset());
-          skipped--;
-        }
+      }
+      return processFetchResponse(fetchResponse, tempTime);
+    } catch (Exception e) {
+      log.info("Exception generated during fetch for topic " + kafkaRequest.getTopic()
+          + ". This topic will be skipped.");
+      e.printStackTrace();
```
HenryCaiHaiying: Don't use e.printStackTrace()

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
```diff
@@ -163,53 +168,91 @@ public boolean fetch() throws IOException {
...
+      return false;
+    }
+  }
 
-        if (!messageIter.hasNext()) {
-          System.out.println("No more data left to process. Returning false");
-          messageIter = null;
-          return false;
+  private void refreshTopicMetadata() {
+    TopicMetadataRequest request = new TopicMetadataRequest(Collections.singletonList(kafkaRequest.getTopic()));
```
HenryCaiHaiying: Should be topic + partition.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
```diff
@@ -163,53 +168,91 @@ public boolean fetch() throws IOException {
...
+      if (partitionMetadata.partitionId() == kafkaRequest.getPartition()) {
+        simpleConsumer = new SimpleConsumer(partitionMetadata.leader().host(), partitionMetadata.leader().port(),
+            CamusJob.getKafkaTimeoutValue(context), CamusJob.getKafkaBufferSize(context),
+            CamusJob.getKafkaClientName(context));
+        break;
+      }
+    }
+  }
+
+  private boolean processFetchResponse(FetchResponse fetchResponse, long tempTime) {
```
HenryCaiHaiying: Is this the same as the old code?  How come it looks like it has more lines.

zliu41: Line 215-249 is the same as old code line 171-205. I added a try-catch outside it, because there's a try catch in old code 207-211.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
zliu41: Fixed. The printStackTrace() is in the old code, and there are other occurrences of printStackTrace() in the old code.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
```diff
@@ -390,6 +390,10 @@ public int compare(FileStatus f1, FileStatus f2) {
         && props.getProperty(ETL_FAIL_ON_ERRORS, Boolean.FALSE.toString()).equalsIgnoreCase(Boolean.TRUE.toString())) {
       throw new RuntimeException("Camus saw errors, check stderr");
     }
+
+    if (EtlInputFormat.reportJobFailureDueToOffsetOutOfRange) {
```
HenryCaiHaiying: It's also better to introduce a config parameter to decide whether to throw Exception on this case.

Also How did you test all these error cases?

zliu41:
1. Added unit test for etl2574: offset out of range
 - added method "testJobFailureDueToOffsetOutOfRange()" in CamusJobTest.java
 - added class "EtlInputFormatFake", the difference from "EtlInputFormat" is that the fake class always return out-of-range offsets.
 - added field "useFakeEtlInputFormatForUnitTest" in CamusJob class. If ==true, it will load EtlInputFormatFake instead of EtlInputFormat.

2. Added unit test for etl2573: fetch() failure
 - Use easymock for kafkaReader.simpleConsumer

3. Added config parameter "etl.fail.on.offset.outofrange".

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
```diff
@@ -163,53 +168,96 @@ public boolean fetch() throws IOException {
     try {
       fetchResponse = simpleConsumer.fetch(fetchRequest);
       if (fetchResponse.hasError()) {
+        throw new RuntimeException();
```
HenryCaiHaiying: The exception should have some error messages.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatFake.java
```diff
@@ -0,0 +1,595 @@
+package com.linkedin.camus.etl.kafka.mapred;
```
HenryCaiHaiying: Can you mock the ETLInputFormat?  There are too many code duplications here.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/common/KafkaReaderTest.java
```diff
@@ -0,0 +1,51 @@
...
+    request.setURI(new URI("http://localhost:8888"));
+    this.kafkaReader = new KafkaReader(context, request, 100, 100);
+  }
+
+  @Test
+  public void testFetchFailure() throws SecurityException, NoSuchFieldException,
+    IllegalArgumentException, IllegalAccessException, IOException {
+    SimpleConsumer consumer = EasyMock.createNiceMock(SimpleConsumer.class);
+    EasyMock.expect(consumer.fetch((FetchRequest) EasyMock.anyObject())).andReturn(null);
+    EasyMock.expectLastCall().times(2);
```
HenryCaiHaiying: Why the fetch will be called two times?

zliu41: One for the first attempt, and one for the retry, after refreshing topic metadata.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
zliu41: This exception is solely for consumption in line 174 (so that we don't need to write duplicate code for fetchResponse=null, fetchResponse.hasError(), and fetch() has exception. All three cases will go to line 174). The error message is printed in line 175.

HenryCaiHaiying: I think it's better to put the fetchResponse.error in the exception message so the caller will have more information to print.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
```diff
@@ -163,53 +168,96 @@ public boolean fetch() throws IOException {
...
+      }
+      return processFetchResponse(fetchResponse, tempTime);
+    } catch (Exception e) {
+      log.info("Exception generated during fetch for topic " + kafkaRequest.getTopic()
+          + ". Will refresh topic metadata and retry.");
```
HenryCaiHaiying: The exception 'e' is not printed in the log.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
zliu41: Fixed.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatFake.java
zliu41: Fixed.

zliu41:
Modified unit test for offset-out-of-range. Instead of using a fake EtlInputFormatFake class which has duplicate code, now using a flag (useMockRequestForUnitTest) in EtlInputFormat.java. If flag==true, EtlInputFormat will replace the actual CamusRequest with a mocked request, which contains out-of-range offsets.

Also fixed other minor issues.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
zliu41: Fixed.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/common/KafkaReaderTest.java
HenryCaiHaiying: OK.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTest.java
```diff
@@ -140,6 +148,32 @@ public void runJob() throws Exception {
   }
 
   @Test
+  public void testJobFailureDueToOffsetTooEarly() throws Exception {
+    EtlInputFormat.useMockRequestForUnitTest = true;
+    createMockRequestOffsetTooEarly();
+    try {
+      job.run();
+      fail("Should have thrown RuntimeException due to offset out of range.");
```
HenryCaiHaiying: Do you need to call EasyMock.verify somewhere?

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
HenryCaiHaiying: The new code looks good.

HenryCaiHaiying:
Looks good to be merged.

zliu41:
Rewrote unit test to use the new mock framework.

PR #179, opened 2015-02-26
Add the unit test framework to be able to mock Kafka SimpleConsumer

HenryCaiHaiying:


camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
```diff
@@ -135,16 +135,21 @@ public static void setLogger(Logger log) {
     return topicMetadataList;
   }
 
-  private SimpleConsumer createConsumer(JobContext context, String broker) {
+  private SimpleConsumer createBrokerConsumer(JobContext context, String broker) {
```
zliu41: 1. Why not name it createSimpleConsumer?
2. This method (as well as lots of other methods) doesn't access object state, and can be static. But this may be too much change and not feasible.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
```diff
@@ -51,7 +52,8 @@
   /**
    * Construct using the json representation of the kafka request
    */
-  public KafkaReader(TaskAttemptContext context, EtlRequest request, int clientTimeout, int fetchBufferSize)
+  public KafkaReader(EtlInputFormat inputFormat, TaskAttemptContext context, EtlRequest request,
```
zliu41: 1. Is 5 parameters too much? Maybe use builder pattern?
2. Instead of passing `EtlInputFormat inputformat`, should we pass `Class<? extends InputFormat> inputFormatClass`, then do `inputFormatClass.newInstance().createSimpleConsumer(...)'?

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlRecordReader.java
```diff
@@ -65,13 +66,16 @@
    * @throws IOException
    * @throws InterruptedException
    */
-  public EtlRecordReader(InputSplit split, TaskAttemptContext context) throws IOException, InterruptedException {
+  public EtlRecordReader(EtlInputFormat inputFormat, InputSplit split, TaskAttemptContext context)
```
zliu41: Same thing: I think inputFormat can be replaced by Class<? extends InputFormat> inputFormatClass. This way, you don't need line 36 and 71. Also it's safer since there's no chance to accidentally modify inputFormat.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -0,0 +1,383 @@
...
+    props.setProperty("mapreduce.framework.name", "local");
+    // Run M/R for Hadoop1
+    props.setProperty("mapreduce.jobtracker.address", "local");
+
+    job = new CamusJob(props);
+  }
+
+  @After
+  public void after() throws IOException {
+    EasyMock.verify(_mocks.toArray());
```
zliu41: Is calling verify() in After method a good practice? What if there's a test that doesn't need to verify all these mocks?

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -0,0 +1,383 @@
...
+    // it to re-instantiate the object with the appropriate output path
+
+    Field field = EtlMultiOutputFormat.class.getDeclaredField("committer");
+    field.setAccessible(true);
+    field.set(null, null);
+  }
+
+  // Kafka Scala layer only provides desearilization as a way to create response, it's very hairy on 
+  // the exact order of those bytes.
+  private static TopicMetadataResponse createTopicMetadataResponseFromBytes(TopicMetadataRequest request) {
```
zliu41: Can you add a boolean parameter to control response.hasError()=true vs false? I may need that.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
HenryCaiHaiying: I do have a createSimpleConsumer() method later in the file.

The createBrokerConsumer is just a simple wrapper to call createSimpleConsumer with broker information.

The createSimpleConsumer() was first created to be an instance method so the subclass can override, the first iteration of the framework was using subclass: TestEtlInputFormat, we might go back to that route.  For this reason, I think it's better to keep it as instance method.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/KafkaReader.java
HenryCaiHaiying: KafkaReader's old constructor already takes 4 arguments, adding one more doesn't seem a lot.  If you know how to get the InputFormat from context, then I don't need this extra argument.

For suggestion #2, I cannot use InputFormat, since this is a hadoop class which doesn't define method: createSimpleConsumer

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlRecordReader.java
HenryCaiHaiying: But I need the createSimpleConsumer method from EtlInputFormat

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
HenryCaiHaiying: I can move it to verifyJob1().

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
HenryCaiHaiying: But then how are you going to set that boolean argument?  So far the design was open the setupJobX() method for each test case to let each test case to explicitly setup the desired behavior for each mock interaction.

If we want to generalize this approach, we would have to come up with the data structure to represent all test data scenarios.

zliu41:
LGTM

PR #180, opened 2015-02-27
Fixed the cast from TaskAttemptContext to Mapper.Context.

zliu41:
In Hadoop 1, TaskAttemptContext.getCounter() is not available, has to cast context to Mapper.Context
However, in Hadoop 2, TaskAttemptContext becomes an interface, and its implementation, TaskAttemptContextImpl, cannot be cast to Mapper.Context.
Using reflection to make it work for both Hadoop 1 and 2.
Regression test passed on Nertz.

### DIFF
--- a/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlMultiOutputRecordWriter.java
+++ b/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlMultiOutputRecordWriter.java
@@ -80,6 +80,7 @@ public class EtlMultiOutputRecordWriter extends RecordWriter<EtlKey, Object> {
           ((Mapper.Context)context).getCounter("total", "skip-old").increment(1);
         } catch (Exception e) {
           log.error("error incrementing counter 'total:skip-old': " + e.getMessage());
+          throw new RuntimeException(e);
         }
         committer.addOffset(key);
       } else {
--- a/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlMultiOutputRecordWriter.java
+++ b/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlMultiOutputRecordWriter.java
@@ -2,12 +2,14 @@ package com.linkedin.camus.etl.kafka.mapred;
 
 import java.io.IOException;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
 import java.util.HashMap;
 
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.SequenceFile;
 import org.apache.hadoop.io.SequenceFile.Writer;
+import org.apache.hadoop.mapreduce.Counter;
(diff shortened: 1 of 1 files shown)


camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlMultiOutputRecordWriter.java
```diff
@@ -68,7 +70,17 @@ public void write(EtlKey key, Object val) throws IOException, InterruptedExcepti
...
-        ((Mapper.Context)context).getCounter("total", "skip-old").increment(1);
+        try {
+          //In Hadoop 2, TaskAttemptContext.getCounter() is available
+          //In Hadoop 2, TaskAttemptContextImpl cannot be cast to Mapper.Context
+          Method getCounterMethod = context.getClass().getMethod("getCounter", String.class, String.class);
+          ((Counter) getCounterMethod.invoke(context, "total", "skip-old")).increment(1);
+        } catch (NoSuchMethodException e) {
+          //In Hadoop 1, TaskAttemptContext.getCounter() is not available, has to cast context to Mapper.Context
+          ((Mapper.Context)context).getCounter("total", "skip-old").increment(1);
+        } catch (Exception e) {
```
HenryCaiHaiying: Does the instanceOfInterface mechanism work?

Why do we need to catch Exception here?  If we catch Exception here for hadoop2, we would need to do the same for hadoop1.

The old code doesn't capture exception, it's actually good in the sense this error wasn't lost in the map log.  If we capture the exception, we would need to record how many times we see the error and fail the job at the end.

zliu41: Now fail the job if we catch exception in getCounterMethod.invoke(), which could be IllegalAccessException, IllegalArgumentException, InvocationTargetException, or RuntimeException. This shouldn't happen unless the accessibility or signature of getCounter() is changed.

zliu41:
Changed catch(Exception e) for getCounterMethod.invoke() to specific exceptions

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlMultiOutputRecordWriter.java
```diff
@@ -68,7 +71,24 @@ public void write(EtlKey key, Object val) throws IOException, InterruptedExcepti
         //TODO: fix this logging message, should be logged once as a total count of old records skipped for each topic
         // for now, commenting this out
         //log.warn("Key's time: " + key + " is less than beginTime: " + beginTimeStamp);
-        ((Mapper.Context)context).getCounter("total", "skip-old").increment(1);
+        try {
```
chavdar: Doing reflection on every event is going to be very expensive. We need to cache the counter objects once at the beginning.

HenryCaiHaiying: Good suggestion, since we know the two runtime classes (between hadoop1 and hadoop2), we can prefetch the method object upfront.

PR #181, opened 2015-02-27
ETL-2564: Error handling in camus: Exception on TopicMetadata call

HenryCaiHaiying:
Deal with the case when the call to retrieve on topic metadata failed, retry three times.

Added two unit tests for the new behavior (one positive, one negative)

### DIFF
--- a/camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatTest.java
+++ b/camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatTest.java
@@ -68,9 +68,9 @@ public class EtlInputFormatTest {
 
     SimpleConsumer simpleConsumer = EasyMock.createMock(SimpleConsumer.class);
     mocks.add(simpleConsumer);
-    EasyMock.expect(simpleConsumer.clientId()).andReturn(DUMMY_VALUE).times(4);
+    EasyMock.expect(simpleConsumer.clientId()).andReturn(DUMMY_VALUE).times(EtlInputFormat.NUM_TRIES_TOPIC_METADATA + 1);
     Exception ex = new RuntimeException("No TopicMeta");
-    EasyMock.expect(simpleConsumer.send((TopicMetadataRequest)EasyMock.anyObject())).andThrow(ex).times(3);
+    EasyMock.expect(simpleConsumer.send((TopicMetadataRequest)EasyMock.anyObject())).andThrow(ex).times(EtlInputFormat.NUM_TRIES_TOPIC_METADATA);
     simpleConsumer.close();
     EasyMock.expectLastCall().andVoid().anyTimes();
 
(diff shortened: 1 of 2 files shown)


camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatTest.java
```diff
@@ -0,0 +1,100 @@
...
+    
+    EtlInputFormat inputFormat = new TestEtlInputFormat(simpleConsumer);    
+    List<TopicMetadata> actualTopicMetadatas = inputFormat.getKafkaMetadata(jobContext);
+    
+    EasyMock.verify(mocks.toArray());
+    
+    assertEquals(actualTopicMetadatas, topicMetadatas);
+  }
+
+  @Test(expected = RuntimeException.class)
```
zliu41: Should verify the exception message? RuntimeException can be thrown from lots of places.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    EasyMock.replay(mocks.toArray());
+
+    EtlInputFormat inputFormat = new TestEtlInputFormat(simpleConsumer);    
+    List<TopicMetadata> actualTopicMetadatas = inputFormat.getKafkaMetadata(jobContext);
+
+    EasyMock.verify(mocks.toArray());
+  }
+  
+  private static class TestEtlInputFormat extends EtlInputFormat {
```
zliu41: This class is already in CamusJobTestWithMock.java. Can you reuse it?

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatTest.java
HenryCaiHaiying: This will be a little difficult.  The exception thrown from EtlInputFormat is not the original exception, it's a wrapped one.

If I code my test to expect a wrapped exception, it will be tightly coupled with the system code, the test will become too brittle to break if the system code changes that wrapping behavior.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatTest.java
HenryCaiHaiying: I can move that class out to reuse it.

camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormatTest.java
zliu41: If you throw new RuntimeException("Failed to obtain metadata!", savedException),
then the exception message should just be "Failed to obtain metadata!", right?
So can you just catch it, and do assertEqual(e.getMessage(), "Failed to obtain metadata!") ?

zliu41:
LGTM

PR #183, opened 2015-02-27
cache counter in EtlMultiOutputRecordWriter to avoid too many reflections

zliu41:
cache the counter as a static variable "private static Counter topicSkipOldCounter". It is initialized in the constructor of EtlMultiOutputRecordWriter.


camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlMultiOutputRecordWriter.java
```diff
@@ -30,6 +30,7 @@
   private String currentTopic = "";
   private long beginTimeStamp = 0;
   private static Logger log = Logger.getLogger(EtlMultiOutputRecordWriter.class);
+  private static Counter topicSkipOldCounter = null;
```
HenryCaiHaiying: I think this can be defined as 'private final' instance variable, so it's guaranteed to be there, there is no need to do a lazy-initialization.

zliu41: Changed to private final.

HenryCaiHaiying:
Thanks.

PR #184, opened 2015-03-02
fix build failures

ydailinkedin:


camus-api/pom.xml
```diff
@@ -1,23 +1,35 @@
...
+		</dependency>
+		<dependency>
+			<groupId>org.apache.hadoop</groupId>
+			<artifactId>hadoop-common</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.easymock</groupId>
+			<artifactId>easymock</artifactId>
+		</dependency>
+	</dependencies>
```
HenryCaiHaiying: Two space indentation should be better.

ydailinkedin: Changed to two space indentations.

PR #185, opened 2015-03-03
add dependency and remove wrong unittest

ydailinkedin:


camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/CamusJobTestWithMock.java
```diff
@@ -274,7 +274,6 @@ public void testJobFailDueToOffsetRangeCallError() throws Exception {
     setupJobFailDueToOffsetRangeCallError();
     job = new CamusJob(props);
     job.run(EtlInputFormatForUnitTest.class, EtlMultiOutputFormat.class);
-    verifyJobSucceed();
```
HenryCaiHaiying: Why is this removed?

zliu41: This job is supposed to fail, so should not verify job succeed. 
This somehow passes on maven, but fails on ligradle.

HenryCaiHaiying: OK, I see.

PR #186, opened 2015-03-03
fix unit test

zliu41:
The after() method in CamusJobTest does not clean up properly, which caused problem for CamusJobTestWithMock.


HenryCaiHaiying:
looks good.

PR #187, opened 2015-03-04
Fix cast issue when getting counter from TaskAttemptContext

zliu41:
ligradle clean build passed.


camus-etl-kafka/src/test/java/com/linkedin/camus/etl/kafka/common/KafkaReaderTest.java
```diff
@@ -27,7 +26,17 @@
   public void before() throws Exception {
     Configuration conf = new Configuration();
     conf.set(CamusJob.KAFKA_CLIENT_NAME, "DummyClientName");
-    TaskAttemptContext context = new TaskAttemptContextImpl(conf, new TaskAttemptID());
+    TaskAttemptContext context = null;
```
zliu41: Original code doesn't build against Hadoop 1 because there's no TaskAttemptContextImpl in hadoop 1. Have to use reflection.

HenryCaiHaiying:
Thanks.

PR #197, opened 2015-03-26
fail job if creating kafka producer failed when sending counds

zliu41:
Currently job fails if producer.send() fails, but sometimes creating the producer itself may fail, and we should also fail the job in this case.

### DIFF
--- a/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/EtlCounts.java
+++ b/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/EtlCounts.java
@@ -206,7 +206,7 @@ public class EtlCounts {
       encoder.init(props, "TrackingMonitoringEvent");
       monitoringDetails =
           (AbstractMonitoringEvent) Class.forName(getMonitoringEventClass(conf))
-          .getDeclaredConstructor(Configuration.class).newInstance(conf);
+              .getDeclaredConstructor(Configuration.class).newInstance(conf);
     } catch (Exception e1) {
       throw new RuntimeException(e1);
     }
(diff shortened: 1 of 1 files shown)


camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/EtlCounts.java
```diff
@@ -275,6 +276,8 @@ private void produceCount(String brokerList, ArrayList<byte[]> monitorSet) {
           }
         }
       }
+    } catch (Exception e) {
+      throw new RuntimeException("failed to publish counts to kafka: " + e.getMessage());
```
sahilTakiar: You probably want to change this to:

throw new RuntimeException("failed to publish counts to Kafka: " + e.getMessage(), e);

So you get the full stack trace of the original exception

zliu41: Fixed.

sahilTakiar:
LGTM. I think that it is still worth taking some time to research or talk to the Kafka team about how exceptions thrown in Scala get propagated to Java. It looks like Scala doesn't have checked exceptions, which is why the constructor doesn't declare that it throws any exception. However, I would expect an unchecked exceptions thrown by Scala code to be treated as a RuntimeException by the JVM and cause the program to crash. If not, then we need to be extra careful when dealing with Kafka code written in Scala.

HenryCaiHaiying:
As I understand, Scala doesn't require the exceptions to be declared on method signature.  The exception thrown from scala is actually checked exception (e.g. IOException), unless you look deep into the kafka scala code, you won't really know what kind of exception will be thrown.  The next version of kafka is going to change back to Java, the exception handling would be more straightforward then.

sahilTakiar:
Still if the constructor throws an IOException, why does the Camus code need to explicitly "catch (Exception e)" and then re-throw is as a RuntimException. If the IOException is not caught then it should still fail the java process.

Is the new consumer Kafka is developing still use Scala?

PR #198, opened 2015-03-31
make camus daily compaction work for hourly

zliu41:
- Use current daily compaction approach for hourly compaction
- If outliers are detected, rather than reprocess the topic, simply add it to outlier folder
- Refactor the daily compaction code and remove unnecessary logging
- Report per-topic metrics for hourly compaction, such as data size, MR submit time, MR duration, etc.

### DIFF
--- a/camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
+++ b/camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
@@ -1,0 +1,327 @@
+package com.linkedin.camus.sweeper;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.Properties;
+import java.util.Map.Entry;
+import java.util.concurrent.Future;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
+import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat;
+import org.apache.hadoop.util.ToolRunner;
+import org.apache.log4j.Logger;
+
+import com.linkedin.camus.sweeper.mapreduce.CamusSweeperJob;
+import com.linkedin.camus.sweeper.utils.Utils;
+
+
+public class CamusHourlySweeper extends CamusSweeper {
+
+  private static final Logger LOG = Logger.getLogger(CamusHourlySweeper.class);
+
+  private static final String REDUCE_COUNT_OVERRIDE = "reduce.count.override.";
+  private static final String CAMUS_SWEEPER_TMP_DIR = "camus.sweeper.tmp.dir";
+  private static final String MAX_FILES = "max.files";
+  private static final int DEFAULT_MAX_FILES = 24;
+  private static final String TMP_PATH = "tmp.path";
+  private static final String CAMUS_SWEEPER_IO_CONFIGURER_CLASS = "camus.sweeper.io.configurer.class";
+  private static final String MAPRED_COMPRESS_MAP_OUTPUT = "mapred.compress.map.output";
+  private static final boolean DEFAULT_MAPRED_COMPRESS_MAP_OUTPUT = Boolean.TRUE;
+  private static final String REDUCER_COUNT = "reducer.count";
+  private static final int DEFAULT_REDUCER_COUNT = 45;
+  private static final String MAPRED_MIN_SPLIT_SIZE = "mapred.min.split.size";
+  private static final String MAPRED_MAX_SPLIT_SIZE = "mapred.max.split.size";
+
+  static final String TOPIC_AND_HOUR = "topic.and.hour";
+  static final String INPUT_PATHS = "input.paths";
+  static final String DEST_PATH = "dest.path";
+
+  private final CamusSweeperMetrics metrics;
+
+  public CamusHourlySweeper() {
+    super();
+    metrics = new CamusSweeperMetrics();
+  }
+
+  public CamusHourlySweeper(Properties props) {
+    super(props);
+    metrics = new CamusSweeperMetrics();
+  }
+
+  @Override
+  public Map<FileStatus, String> findAllTopics(Path input, PathFilter filter, String topicSubdir, FileSystem fs)
+      throws IOException {
+    Map<FileStatus, String> topics = new HashMap<FileStatus, String>();
+    for (FileStatus f : fs.listStatus(input)) {
+      if (f.isDir() && filter.accept(f.getPath())) {
+        LOG.info("found topic: " + f.getPath().getName());
+        topics.put(fs.getFileStatus(f.getPath()), f.getPath().getName());
+      }
+    }
+    return topics;
+  }
+
+  @Override
+  public void run() throws Exception {
+    metrics.timeStart = System.currentTimeMillis();
+    super.run();
+    reportMetrics();
+    addOutliers();
+  }
+
+  private void addOutliers() throws IOException {
+    Configuration conf = new Configuration();
+    for (Entry<Object, Object> pair : props.entrySet()) {
+      String key = (String) pair.getKey();
+      conf.set(key, (String) pair.getValue());
+    }
+    for (Properties jobProps : planner.getOutlierProperties()) {
+      this.executorService.submit(new OutlierCollectorRunner(jobProps, FileSystem.get(conf)));
+    }
+  }
+
+  private void reportMetrics() throws IllegalArgumentException, IOException {
+    LOG.info("reporting metrics");
+    metrics.reportTotalRunningTime();
+    metrics.reportTotalDataSize();
+    metrics.reportDataSizeByTopic();
+    metrics.reportDurationFromStartToRunnerStart();
+    metrics.reportDurationFromRunnerStartToMRSubmitted();
+    metrics.reportDurationFromMRSubmittedToMRStarted();
+    metrics.reportDurationFromMRStartedToMRFinished();
+    LOG.info("finished reporting metrics");
+  }
+
+  public static void main(String args[]) throws Exception {
+    CamusSweeper job = new CamusHourlySweeper();
+    ToolRunner.run(job, args);
+  }
+
+  @Override
+  protected Future<?> runCollector(Properties props, String topic) {
+    String jobName = topic + "-" + UUID.randomUUID().toString();
+    props.put("tmp.path", props.getProperty(CAMUS_SWEEPER_TMP_DIR) + "/" + jobName + "_" + System.currentTimeMillis());
+
+    if (props.containsKey(REDUCE_COUNT_OVERRIDE + topic))
+      props.put("reducer.count", Integer.parseInt(props.getProperty(REDUCE_COUNT_OVERRIDE + topic)));
+
+    LOG.info("Processing " + props.get(INPUT_PATHS));
+
+    return executorService.submit(new KafkaCollectorRunner(jobName, props, errorMessages, topic));
+  }
+
+  @Override
+  protected void runCollectorForTopicDir(FileSystem fs, String topic, Path topicSourceDir, Path topicDestDir)
+      throws Exception {
+    LOG.info("Running collector for topic " + topic + " source:" + topicSourceDir + " dest:" + topicDestDir);
+    List<Future<?>> tasksToComplete = new ArrayList<Future<?>>();
+
+    List<Properties> jobPropsList = planner.createSweeperJobProps(topic, topicSourceDir, topicDestDir, fs, metrics);
+
+    for (Properties jobProps : jobPropsList) {
+      tasksToComplete.add(runCollector(jobProps, topic));
+    }
+  }
+
+  private class KafkaCollectorRunner extends CamusSweeper.KafkaCollectorRunner {
+
+    public KafkaCollectorRunner(String name, Properties props, List<SweeperError> errorQueue, String topic) {
+      super(name, props, errorQueue, topic);
+    }
+
+    @Override
+    public void run() {
+      KafkaCollector collector = null;
+      try {
+        LOG.info("Starting runner for " + this.props.getProperty(TOPIC_AND_HOUR));
+        collector = new KafkaCollector(props, name, topic);
+        LOG.info("Running " + name + " for input " + props.getProperty(INPUT_PATHS));
+        collector.run();
+      } catch (Throwable e) { // Sometimes the error is the Throwable, e.g. java.lang.NoClassDefFoundError
+        e.printStackTrace();
+        LOG.error("Failed for " + name + " ,job: " + collector == null ? null : collector.getJob() + " failed for "
+            + props.getProperty(INPUT_PATHS) + " Exception:" + e.getLocalizedMessage());
+        errorQueue.add(new SweeperError(name, props.get(INPUT_PATHS).toString(), e));
+      }
+    }
+  }
+
+  private class KafkaCollector extends CamusSweeper.KafkaCollector {
+
+    private final String topicAndHour;
+    private final Path[] inputPaths;
+    private final Path tmpPath;
+    private final Path outputPath;
+    private final FileSystem fs;
+
+    public KafkaCollector(Properties props, String jobName, String topicName) throws IOException {
+      super(props, jobName, topicName);
+      this.topicAndHour = props.getProperty(TOPIC_AND_HOUR);
+      this.fs = FileSystem.get(job.getConfiguration());
+      this.inputPaths = getInputPaths();
+      this.tmpPath = new Path(job.getConfiguration().get(TMP_PATH));
+      this.outputPath = new Path(job.getConfiguration().get(DEST_PATH));
+      addInputAndOutputPathsToFileInputFormat();
+    }
+
+    private Path[] getInputPaths() {
+      List<String> strPaths = Utils.getStringList(props, INPUT_PATHS);
+      Path[] inputPaths = new Path[strPaths.size()];
+      for (int i = 0; i < strPaths.size(); i++)
+        inputPaths[i] = new Path(strPaths.get(i));
+      return inputPaths;
+    }
+
+    @Override
+    public void run() throws Exception {
+      CamusHourlySweeper.this.metrics.runnerStartTimeByTopic.put(this.topicAndHour, System.currentTimeMillis());
+
+      job.getConfiguration().setBoolean(MAPRED_COMPRESS_MAP_OUTPUT, DEFAULT_MAPRED_COMPRESS_MAP_OUTPUT);
+
+      ((CamusSweeperJob) Class.forName(props.getProperty(CAMUS_SWEEPER_IO_CONFIGURER_CLASS)).newInstance()).setLogger(
+          LOG).configureJob(topicName, job);
+
+      setNumOfReducersAndSplitSizes();
+      submitMrJob();
+      moveTmpPathToOutputPath();
+
+      CamusHourlySweeper.this.metrics.mrFinishTimeByTopic.put(this.topicAndHour, System.currentTimeMillis());
+    }
+
+    private void moveTmpPathToOutputPath() throws IOException {
+      Path oldPath = null;
+      if (fs.exists(outputPath)) {
+        oldPath = new Path("/tmp", "_old_" + job.getJobID());
+        moveExistingContentInOutputPathToOldPath(oldPath);
+      }
+
+      LOG.info("Moving " + tmpPath + " to " + outputPath);
+      mkdirs(fs, outputPath.getParent(), perm);
+
+      if (!fs.rename(tmpPath, outputPath)) {
+        fs.rename(oldPath, outputPath);
+        fs.delete(tmpPath, true);
+        throw new RuntimeException("Error: cannot rename " + tmpPath + " to " + outputPath);
+      }
+      deleteOldPath(oldPath);
+    }
+
+    private void deleteOldPath(Path oldPath) throws IOException {
+      if (oldPath != null && fs.exists(oldPath)) {
+        LOG.info("Deleting " + oldPath);
+        fs.delete(oldPath, true);
+      }
+    }
+
+    private void moveExistingContentInOutputPathToOldPath(Path oldPath) throws IOException {
+      LOG.info("Path " + outputPath + " exists. Overwriting.");
+      if (!fs.rename(outputPath, oldPath)) {
+        fs.delete(tmpPath, true);
+        throw new RuntimeException("Error: cannot rename " + outputPath + " to " + oldPath);
+      }
+    }
+
+    private void setNumOfReducersAndSplitSizes() throws IOException {
+      long inputSize = getInputSize();
+
+      int maxFiles = job.getConfiguration().getInt(MAX_FILES, DEFAULT_MAX_FILES);
+      int numTasks = Math.min((int) (inputSize / targetFileSize) + 1, maxFiles);
+
+      if (job.getNumReduceTasks() != 0) {
+        determineAndSetNumOfReducers(numTasks);
+      } else {
+        setSplitSizes(inputSize / numTasks);
+      }
+    }
+
+    private void submitMrJob() throws IOException, InterruptedException, ClassNotFoundException {
+      CamusHourlySweeper.this.metrics.mrSubmitTimeByTopic.put(this.topicAndHour, System.currentTimeMillis());
+      job.submit();
+      runningJobs.add(job);
+
+      CamusHourlySweeper.this.metrics.mrStartRunningTimeByTopic.put(this.topicAndHour, System.currentTimeMillis());
+
+      LOG.info("job running for: " + job.getConfiguration().get(TOPIC_AND_HOUR) + ", url: " + job.getTrackingURL());
+      job.waitForCompletion(false);
+      if (!job.isSuccessful()) {
+        throw new RuntimeException("hadoop job failed.");
+      }
+    }
+
+    private void setSplitSizes(long targetSplitSize) {
+      LOG.info("Setting target split size " + targetSplitSize);
+      job.getConfiguration().setLong(MAPRED_MAX_SPLIT_SIZE, targetSplitSize);
+      job.getConfiguration().setLong(MAPRED_MIN_SPLIT_SIZE, targetSplitSize);
+    }
+
+    private void determineAndSetNumOfReducers(int numTasks) {
+      int numReducers;
+      if (job.getConfiguration().get(REDUCER_COUNT) != null) {
+        numReducers = job.getConfiguration().getInt(REDUCER_COUNT, DEFAULT_REDUCER_COUNT);
+      } else {
+        numReducers = numTasks;
+      }
+      job.setNumReduceTasks(numReducers);
+    }
+
+    private long getInputSize() throws IOException {
+      long inputSize = 0;
+      for (Path p : inputPaths) {
+        LOG.info("inputPath: " + p.toString() + ", size=" + fs.getContentSummary(p).getLength());
+        inputSize += fs.getContentSummary(p).getLength();
+      }
+      return inputSize;
+    }
+
+    private void addInputAndOutputPathsToFileInputFormat() throws IOException {
+      for (Path path : inputPaths) {
+        FileInputFormat.addInputPath(job, path);
+        FileOutputFormat.setOutputPath(job, tmpPath);
+      }
+    }
+  }
+
+  private class OutlierCollectorRunner implements Runnable {
+
+    private final Properties props;
+    private final FileSystem fs;
+
+    public OutlierCollectorRunner(Properties props, FileSystem fs) {
+      this.props = props;
+      this.fs = fs;
+    }
+
+    @Override
+    public void run() {
+      String inputPaths = this.props.getProperty(CamusHourlySweeper.INPUT_PATHS);
+      String outputPathStr = this.props.getProperty(CamusHourlySweeper.DEST_PATH);
+      Path outputPath = new Path(outputPathStr, "outlier");
+
+      try {
+        fs.mkdirs(outputPath);
+        long destinationModTime = fs.getFileStatus(new Path(outputPathStr)).getModificationTime();
+
+        for (String inputPathStr : inputPaths.split(",")) {
+          Path inputPath = new Path(inputPathStr);
+          for (FileStatus status : fs.globStatus(new Path(inputPath, "*"), new HiddenFilter())) {
+            if (status.getModificationTime() > destinationModTime) {
+              LOG.info("moving " + status.getPath() + " to " + outputPath);
+              fs.rename(status.getPath(), new Path(outputPath, status.getPath().getName()));
+            }
+          }
+        }
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    }
+  }
+}
--- a/camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
+++ b/camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
@@ -187,8 +187,6 @@ public class CamusHourlySweeper extends CamusSweeper {
       setNumOfReducersAndSplitSizes();
       submitMrJob();
       moveTmpPathToOutputPath();
-
-      CamusHourlySweeper.this.metrics.recordMrFinishTimeByTopic(this.topicAndHour, System.currentTimeMillis());
     }
 
     @Override
@@ -199,8 +197,9 @@ public class CamusHourlySweeper extends CamusSweeper {
 
       CamusHourlySweeper.this.metrics.recordMrStartRunningTimeByTopic(this.topicAndHour, System.currentTimeMillis());
 
-      LOG.info("job running for: " + job.getConfiguration().get(TOPIC_AND_HOUR) + ", url: " + job.getTrackingURL());
+      LOG.info("job running for: " + props.getProperty(TOPIC_AND_HOUR) + ", url: " + job.getTrackingURL());
       job.waitForCompletion(false);
+      CamusHourlySweeper.this.metrics.recordMrFinishTimeByTopic(this.topicAndHour, System.currentTimeMillis());
       if (!job.isSuccessful()) {
         throw new RuntimeException("hadoop job failed.");
       }
--- a/camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
+++ b/camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
@@ -25,9 +25,6 @@ public class CamusHourlySweeper extends CamusSweeper {
 
   private static final Logger LOG = Logger.getLogger(CamusHourlySweeper.class);
 
-  private static final String REDUCE_COUNT_OVERRIDE = "reduce.count.override.";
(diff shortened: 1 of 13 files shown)


camus-sweeper/src/main/java/com/linkedin/camus/sweeper/HiddenFilter.java
```diff
@@ -0,0 +1,17 @@
...
+public class HiddenFilter implements PathFilter {
+  @Override
+  public boolean accept(Path path) {
+    String name = path.getName();
+    if (name.startsWith("_") || name.startsWith(".")) {
+      return false;
+    }
+
+    return true;
+  }
```
sahilTakiar: You can replace this with

return !(name.startsWith("_") || name.startsWith("."))

Also can you add java docs for this class

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+    metrics = new CamusSweeperMetrics();
+  }
+
+  public CamusHourlySweeper(Properties props) {
+    super(props);
+    metrics = new CamusSweeperMetrics();
+  }
+
+  @Override
+  public Map<FileStatus, String> findAllTopics(Path input, PathFilter filter, String topicSubdir, FileSystem fs)
```
sahilTakiar: Does the code actually require tracking the entire FileStatus object as the key, or can you just use the file Path as the key instead?

sahilTakiar:
Can we add some more javadocs, or is there an existing design wiki? It's not clear what each class does unless you already the know the Camus code base.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+    private final FileSystem fs;
+
+    public KafkaCollector(Properties props, String jobName, String topicName) throws IOException {
+      super(props, jobName, topicName);
+      this.topicAndHour = props.getProperty(TOPIC_AND_HOUR);
+      this.fs = FileSystem.get(job.getConfiguration());
+      this.inputPaths = getInputPaths();
+      this.tmpPath = new Path(job.getConfiguration().get(TMP_PATH));
+      this.outputPath = new Path(job.getConfiguration().get(DEST_PATH));
+      addInputAndOutputPathsToFileInputFormat();
```
sahilTakiar: Makes more sense to do props.getProperty("prop_name") rather than job.getConfiguration().get("prop_name")

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+
+  public static void main(String args[]) throws Exception {
+    CamusSweeper job = new CamusHourlySweeper();
+    ToolRunner.run(job, args);
+  }
+
+  @Override
+  protected Future<?> runCollector(Properties props, String topic) {
+    String jobName = topic + "-" + UUID.randomUUID().toString();
+    props.put("tmp.path", props.getProperty(CAMUS_SWEEPER_TMP_DIR) + "/" + jobName + "_" + System.currentTimeMillis());
```
sahilTakiar: Use the static final variable initialized in the constructor TMP_PATH rather than hard-coding the string here.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+
+    if (props.containsKey(REDUCE_COUNT_OVERRIDE + topic))
+      props.put("reducer.count", Integer.parseInt(props.getProperty(REDUCE_COUNT_OVERRIDE + topic)));
```
sahilTakiar: Same as above, create a static final string for this config

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+      List<String> strPaths = Utils.getStringList(props, INPUT_PATHS);
+      Path[] inputPaths = new Path[strPaths.size()];
+      for (int i = 0; i < strPaths.size(); i++)
+        inputPaths[i] = new Path(strPaths.get(i));
+      return inputPaths;
+    }
+
+    @Override
+    public void run() throws Exception {
+      CamusHourlySweeper.this.metrics.runnerStartTimeByTopic.put(this.topicAndHour, System.currentTimeMillis());
```
sahilTakiar: Rather than directly accessing these instance variables from the metrics class, it would be clearer if you have a method such as: reportHourlyRunnerStartTime(String topic, String hour, long startTime)

sahilTakiar:
Some unit tests would be good too considering there is a far amount of code change. This feature will probably continue to be used even when the new Kafka consumer code is written, and even after the migration to Gobblin is done.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+        LOG.info("inputPath: " + p.toString() + ", size=" + fs.getContentSummary(p).getLength());
+        inputSize += fs.getContentSummary(p).getLength();
+      }
+      return inputSize;
+    }
+
+    private void addInputAndOutputPathsToFileInputFormat() throws IOException {
+      for (Path path : inputPaths) {
+        FileInputFormat.addInputPath(job, path);
+        FileOutputFormat.setOutputPath(job, tmpPath);
```
sahilTakiar: You can move this line outside the for loop

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+    private void deleteOldPath(Path oldPath) throws IOException {
+      if (oldPath != null && fs.exists(oldPath)) {
+        LOG.info("Deleting " + oldPath);
+        fs.delete(oldPath, true);
+      }
+    }
+
+    private void moveExistingContentInOutputPathToOldPath(Path oldPath) throws IOException {
+      LOG.info("Path " + outputPath + " exists. Overwriting.");
+      if (!fs.rename(outputPath, oldPath)) {
```
sahilTakiar: It's not actually over-writing data is it? Its moving the data to a tmp location.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+      moveTmpPathToOutputPath();
+
+      CamusHourlySweeper.this.metrics.mrFinishTimeByTopic.put(this.topicAndHour, System.currentTimeMillis());
+    }
+
+    private void moveTmpPathToOutputPath() throws IOException {
+      Path oldPath = null;
+      if (fs.exists(outputPath)) {
+        oldPath = new Path("/tmp", "_old_" + job.getJobID());
+        moveExistingContentInOutputPathToOldPath(oldPath);
```
sahilTakiar: Is this the best approach. If the output data already exists, then it is being moved to some tmp folder on HDFS. Are there any race conditions where the output path could already exist, and the job mistaking-enly moves legitimate data to some random output folder. 

Also, why does the oldPath need to be in the /tmp/ folder? Data in /tmp/ can get deleted by hadoop-ops anytime if the quota for /tmp/ fills up. Is there retention on the data that gets moved into /tmp/? Also, since this is open-source not all users may want to move data to /tmp/.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+      CamusHourlySweeper.this.metrics.mrStartRunningTimeByTopic.put(this.topicAndHour, System.currentTimeMillis());
+
+      LOG.info("job running for: " + job.getConfiguration().get(TOPIC_AND_HOUR) + ", url: " + job.getTrackingURL());
+      job.waitForCompletion(false);
+      if (!job.isSuccessful()) {
+        throw new RuntimeException("hadoop job failed.");
+      }
+    }
+
+    private void setSplitSizes(long targetSplitSize) {
```
sahilTakiar: By setting the split size you are controlling the number of map tasks that get launched. You may want to add that to the javadocs for this method so that it is clear this code path determines the # of map tasks that get launched.

Also, is there any other way to control the number of map tasks that get launched? What if we want to throttle the hourly compaction job because it is taking up too many resources.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+        LOG.info("found topic: " + f.getPath().getName());
+        topics.put(fs.getFileStatus(f.getPath()), f.getPath().getName());
+      }
+    }
+    return topics;
+  }
+
+  @Override
+  public void run() throws Exception {
+    metrics.timeStart = System.currentTimeMillis();
```
liyinan926: Better to create a setter for timeStart.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+    super.run();
+    reportMetrics();
+    addOutliers();
+  }
+
+  private void addOutliers() throws IOException {
+    Configuration conf = new Configuration();
```
liyinan926: If `CamusSweeper` extends `Configured`, you can get a `Configuration` using `getConf()`.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+    for (Entry<Object, Object> pair : props.entrySet()) {
+      String key = (String) pair.getKey();
+      conf.set(key, (String) pair.getValue());
+    }
+    for (Properties jobProps : planner.getOutlierProperties()) {
+      this.executorService.submit(new OutlierCollectorRunner(jobProps, FileSystem.get(conf)));
```
liyinan926: Who's responsible for shutting down the `executorService`?

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+    LOG.info("Processing " + props.get(INPUT_PATHS));
+
+    return executorService.submit(new KafkaCollectorRunner(jobName, props, errorMessages, topic));
+  }
+
+  @Override
+  protected void runCollectorForTopicDir(FileSystem fs, String topic, Path topicSourceDir, Path topicDestDir)
+      throws Exception {
+    LOG.info("Running collector for topic " + topic + " source:" + topicSourceDir + " dest:" + topicDestDir);
+    List<Future<?>> tasksToComplete = new ArrayList<Future<?>>();
```
sahilTakiar: What is the use of this variable? tasksToComplete is not used anywhere after line 131

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
liyinan926: It's fragile to construct path by concatenating path components, due to absence/presence of the trailing slash. Better to do `new Path(props.getProperty(CAMUS_SWEEPER_TMP_DIR), jobName + "_" + System.currentTimeMillis()).toUri().getPath()`.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+    @Override
+    public void run() {
+      KafkaCollector collector = null;
+      try {
+        LOG.info("Starting runner for " + this.props.getProperty(TOPIC_AND_HOUR));
+        collector = new KafkaCollector(props, name, topic);
+        LOG.info("Running " + name + " for input " + props.getProperty(INPUT_PATHS));
+        collector.run();
+      } catch (Throwable e) { // Sometimes the error is the Throwable, e.g. java.lang.NoClassDefFoundError
+        e.printStackTrace();
```
liyinan926: No printStackTrace() please.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+        LOG.error("Failed for " + name + " ,job: " + collector == null ? null : collector.getJob() + " failed for "
```
liyinan926: Seems better to use String.format() here. Also you can do `LOG.error(msg, e);` so you don't need to include the exception message in `msg`.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+    private final Path tmpPath;
+    private final Path outputPath;
+    private final FileSystem fs;
+
+    public KafkaCollector(Properties props, String jobName, String topicName) throws IOException {
+      super(props, jobName, topicName);
+      this.topicAndHour = props.getProperty(TOPIC_AND_HOUR);
+      this.fs = FileSystem.get(job.getConfiguration());
+      this.inputPaths = getInputPaths();
+      this.tmpPath = new Path(job.getConfiguration().get(TMP_PATH));
```
liyinan926: What if TMP_PATH is not set?

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+      job.getConfiguration().setBoolean(MAPRED_COMPRESS_MAP_OUTPUT, DEFAULT_MAPRED_COMPRESS_MAP_OUTPUT);
+
+      ((CamusSweeperJob) Class.forName(props.getProperty(CAMUS_SWEEPER_IO_CONFIGURER_CLASS)).newInstance()).setLogger(
+          LOG).configureJob(topicName, job);
+
+      setNumOfReducersAndSplitSizes();
+      submitMrJob();
+      moveTmpPathToOutputPath();
+
+      CamusHourlySweeper.this.metrics.mrFinishTimeByTopic.put(this.topicAndHour, System.currentTimeMillis());
```
liyinan926: Do you also want to record this metric even when the job fails?

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+          Path inputPath = new Path(inputPathStr);
+          for (FileStatus status : fs.globStatus(new Path(inputPath, "*"), new HiddenFilter())) {
+            if (status.getModificationTime() > destinationModTime) {
+              LOG.info("moving " + status.getPath() + " to " + outputPath);
+              fs.rename(status.getPath(), new Path(outputPath, status.getPath().getName()));
+            }
+          }
+        }
+      } catch (IOException e) {
+        throw new RuntimeException(e);
```
sahilTakiar: You shouldn't throw RuntimExceptions from a Runnable. I believe the ExecutorService will just log the error, but won't fail and the process will continue on. Instead you should use the Callable interface, which can throw checked exceptions and can return an exit value.

http://stackoverflow.com/questions/141284/the-difference-between-the-runnable-and-callable-interfaces-in-java

Also, to check for errors thrown by runnable / callable you need to query the Future object:

http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html#get%28%29

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+      super(props, jobName, topicName);
+      this.topicAndHour = props.getProperty(TOPIC_AND_HOUR);
+      this.fs = FileSystem.get(job.getConfiguration());
+      this.inputPaths = getInputPaths();
+      this.tmpPath = new Path(job.getConfiguration().get(TMP_PATH));
+      this.outputPath = new Path(job.getConfiguration().get(DEST_PATH));
+      addInputAndOutputPathsToFileInputFormat();
+    }
+
+    private Path[] getInputPaths() {
```
liyinan926: Why not just use the list form of inputPaths?

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+      CamusHourlySweeper.this.metrics.runnerStartTimeByTopic.put(this.topicAndHour, System.currentTimeMillis());
+
+      job.getConfiguration().setBoolean(MAPRED_COMPRESS_MAP_OUTPUT, DEFAULT_MAPRED_COMPRESS_MAP_OUTPUT);
+
+      ((CamusSweeperJob) Class.forName(props.getProperty(CAMUS_SWEEPER_IO_CONFIGURER_CLASS)).newInstance()).setLogger(
+          LOG).configureJob(topicName, job);
+
+      setNumOfReducersAndSplitSizes();
+      submitMrJob();
+      moveTmpPathToOutputPath();
```
liyinan926: Do you need to cleanup the tmp path after the job is done?

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: This is the old code, I just refactored it in the new subclass (CamusHourlySweeper) to break some very long methods in the old code (CamusSweeper) into smaller methods. Now I've put the refactored code in the old class.

The old code has a lot of places that didn't use best-practice. I'm not planning to make any major change to the old code, since Camus is going to retire and it's better not mess with it.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Ditto. This is old code

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Ditto. This is old code

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Ditto. This is old code

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Ditto. This is old code

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Ditto. This is old code

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+        numReducers = numTasks;
+      }
+      job.setNumReduceTasks(numReducers);
+    }
+
+    private long getInputSize() throws IOException {
+      long inputSize = 0;
+      for (Path p : inputPaths) {
+        LOG.info("inputPath: " + p.toString() + ", size=" + fs.getContentSummary(p).getLength());
+        inputSize += fs.getContentSummary(p).getLength();
```
liyinan926: It's more common to use `getFileStatus(p).getLen()`.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Ditto. This is old code

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Ditto. This is old code

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+      }
+      return inputSize;
+    }
+
+    private void addInputAndOutputPathsToFileInputFormat() throws IOException {
+      for (Path path : inputPaths) {
+        FileInputFormat.addInputPath(job, path);
```
liyinan926: There's also `FileInputFormat.setInputPaths()` that allows you to set a collection of paths.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Fixed.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+      String outputPathStr = this.props.getProperty(CamusHourlySweeper.DEST_PATH);
+      Path outputPath = new Path(outputPathStr, "outlier");
+
+      try {
+        fs.mkdirs(outputPath);
+        long destinationModTime = fs.getFileStatus(new Path(outputPathStr)).getModificationTime();
+
+        for (String inputPathStr : inputPaths.split(",")) {
+          Path inputPath = new Path(inputPathStr);
+          for (FileStatus status : fs.globStatus(new Path(inputPath, "*"), new HiddenFilter())) {
```
liyinan926: I don't think you need to use `HiddenFilter` explicitly as it is always used internally.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -0,0 +1,327 @@
...
+    public void run() {
+      String inputPaths = this.props.getProperty(CamusHourlySweeper.INPUT_PATHS);
+      String outputPathStr = this.props.getProperty(CamusHourlySweeper.DEST_PATH);
+      Path outputPath = new Path(outputPathStr, "outlier");
+
+      try {
+        fs.mkdirs(outputPath);
+        long destinationModTime = fs.getFileStatus(new Path(outputPathStr)).getModificationTime();
+
+        for (String inputPathStr : inputPaths.split(",")) {
```
liyinan926: Better to use a `Splitter`.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeperPlanner.java
```diff
@@ -0,0 +1,133 @@
...
+
+public class CamusHourlySweeperPlanner extends CamusSweeperPlanner {
+  private static final String CAMUS_HOURLY_SWEEPER_MAX_HOURS_AGO = "camus.hourly.sweeper.max.hours.ago";
+  private static final String DEFAULT_CAMUS_HOURLY_SWEEPER_MAX_HOURS_AGO = "1";
+  private static final String CAMUS_HOURLY_SWEEPER_MIN_HOURS_AGO = "camus.hourly.sweeper.min.hours.ago";
+  private static final String DEFAULT_CAMUS_HOURLY_SWEEPER_MIN_HOURS_AGO = "1";
+
+  private static final Logger LOG = Logger.getLogger(CamusHourlySweeperPlanner.class);
+
+  private DateTimeFormatter hourFormatter;
```
liyinan926: Both can be final.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeperPlanner.java
```diff
@@ -0,0 +1,133 @@
...
+  private DateUtils dUtils;
+
+  @Override
+  public CamusSweeperPlanner setPropertiesLogger(Properties props, Logger log) {
+    dUtils = new DateUtils(props);
+    hourFormatter = dUtils.getDateTimeFormatter("YYYY/MM/dd/HH");
```
liyinan926: The formatter string can be defined as a constant.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeperPlanner.java
```diff
@@ -0,0 +1,133 @@
...
+    return super.setPropertiesLogger(props, log);
+  }
+
+  private DateTime getFolderHour(Path datePath, Path inputDir) {
+    String datePathStr = datePath.toString();
+    String inputDirStr = inputDir.toString();
+    String dateStr = datePathStr.substring(datePathStr.indexOf(inputDirStr) + inputDirStr.length());
```
liyinan926: What if datePathStr.indexOf() returns -1?

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeperPlanner.java
```diff
@@ -0,0 +1,133 @@
...
+      CamusSweeperMetrics metrics) throws IOException {
+    LOG.info("creating hourly sweeper job props: topic=" + topic + ", inputDir=" + inputDir + ", outputDir="
+        + outputDir);
+    List<Properties> jobPropsList = new ArrayList<Properties>();
+    if (!fs.exists(inputDir)) {
+      LOG.warn("inputdir " + inputDir + " does not exist. Skipping topic " + topic);
+      return jobPropsList;
+    }
+
+    for (FileStatus f : fs.globStatus(new Path(inputDir, "*/*/*/*"))) {
```
liyinan926: Ditto. The string can be defined as a constant variable.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeperPlanner.java
zliu41: Should never happens. The old code does the same thing to get timestamp on folder, except old code gets date, new code gets hour.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeperPlanner.java
zliu41: They are not assigned in constructor, but in method "setPropertiesLogger()". This is the how the old code is written and I'm just overriding it and changing dayFormatter to hourFormatter.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeperPlanner.java
```diff
@@ -0,0 +1,133 @@
...
+      }
+    }
+
+    return jobPropsList;
+  }
+
+  private Properties createJobProps(String topic, Path folder, DateTime folderHour, Path outputDir, FileSystem fs,
+      CamusSweeperMetrics metrics) throws IOException {
+
+    Properties jobProps = new Properties(props);
```
liyinan926: Should create an empty Properties using `new Properties()`. The single parameter is used for default.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Guava is not used in Camus. The old code use String.split() at multiple other places.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: This class is a nested class in the old code, I'm just making it a top-level class. There are other methods in the old code that use this class, which I'm not planning to change.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusSweeperPlanner.java
```diff
@@ -63,15 +69,21 @@ private boolean shouldReprocess(FileSystem fs, Path source, long destinationModT
...
-      String name = path.getName();
-      if (name.startsWith("_") || name.startsWith(".")) {
-        return false;
-      }
+  protected String pathListToCommaSeperated(List<Path> list) {
+    ArrayList<Path> tmpList = new ArrayList<Path>();
+    tmpList.addAll(list);
 
-      return true;
+    StringBuilder sb = new StringBuilder(tmpList.get(0).toString());
```
liyinan926: You can start with an empty StringBuilder so you don't need to do this special logic.

zliu41: This method is moved here from subclass CamusSweeperDatePartitionPlanner, since I need to use it in another subclass CamusHourlySweeprPlanner. I agree this can be refactored, but so are lots of other places in the old Camus code, so I'm not planning to do it.

zliu41:
Added some javadoc in CamusHourlySweeper and CamusHourlySweeperPlanner.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Fixed.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Fixed.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: fixed

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Fixed.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: fixed

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: after rename() tmppath should be gone. The old code (daily sweeper) doesn't clean up tmp path.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/HiddenFilter.java
zliu41: Fixed.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Fixed.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeperPlanner.java
zliu41: Fixed.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeperPlanner.java
zliu41: Fixed.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Doesn't seem to need entire FileStatus, but this is overriding the superclass method, and I'm not planning to refactor superclass, which may create more bugs and need more testing.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Fixed.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Fixed. now record the metric even if job fails.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Ditto. this is old code.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: This is old code. Put it back to the old class (CamusSweeper).

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: This is old code. setInputPaths() takes either Path... inputPaths or String commaSeparatedPaths which is not easy to use here.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeperPlanner.java
zliu41: Fixed.

zliu41:
Added unit test for CamusHourlySweeprPlanner.createSweeperJobProps(), which seems to be the only method that should/can be unit-tested. Other code will be tested during validation test.

liyinan926:
LGTM.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
sahilTakiar: Ok, sounds good. I still think we should make a JIRA for code issues that have come up in this pull request. They are still legitimate concerns that should be fixed. While Camus may eventually get replaced by Gobblin, I don't see the hourly or daily compaction code getting replaced by Gobblin any time soon as Gobblin doesn't have the capability of running reducers.

sahilTakiar:
LGTM

PR #199, opened 2015-04-03
Partial Fix for EOF/Coruption and Copy AND Data Loss issue 189

bmistry13:
| **Property Name**                                     | **Default Value** | **Description**                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
|---------------------------------------------------|---------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| etl.final.file.overwrite                          | false         | This is safely option allow no data loss and rename does not overwrite file (e.g when hadoop job partially commits file or admin issue 'hadoop job -kill job_id' etc and file name get overwritten (This will allow to detect file duplication etc).  **FALSE:** This option will allow Final Destination file cannot be overwritten.,If final Destination Path exists that was generated by Partitioner Interface Class then it is fail the Camus Job.   **TRUE:** This option will allow Final Destination File can be overwritten.                                                                              |
| etl.output.two.steps.commit.copyToTemp.and.rename | false         | For Hadoop system when you have volume based directory(MapR) then you need this option and also avoid File getting access while it is getting written (using temp folder).  There is cost to this and "Total copy-rename-commit time(s): 10" under Timing Report.  **FALSE:**  When camus job goes out of range, it will not be move offset and camus job will continue to skip that partition and fail the job.  **TRUE:** When camus job goes out of range, it will automatically move to latest offset and job will not fail and data will be pulled from this partition.(True should be default in my opinion). |

**How to find the cost of "copy-temp-rename" strategy or having this "etl.output.two.steps.commit.copyToTemp.and.rename=true" ?**

Here is sample camus time report with cost in second "Total copy-rename-commit time(s): 10".  The cost is calculated as hadoop counter and should work with Hadoop 1.x or 2.x API.

FYI:  http://stackoverflow.com/questions/21868430/locking-a-directory-in-hdfs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,7 @@ camus-api/target/
 camus-etl-kafka/target/
 camus-example/target/
 camus-schema-registry/target/
+
+release.properties
+pom.xml.versionsBackup
+pom.xml.releaseBackup
--- a/camus-api/pom.xml
+++ b/camus-api/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>com.linkedin.camus</groupId>
     <artifactId>camus-parent</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
+    <version>0.1.1-SNAPSHOT</version>
   </parent>
 
   <artifactId>camus-api</artifactId>
--- a/camus-api/src/main/java/com/linkedin/camus/etl/RecordWriterProvider.java
+++ b/camus-api/src/main/java/com/linkedin/camus/etl/RecordWriterProvider.java
@@ -1,10 +1,11 @@
 package com.linkedin.camus.etl;
 
-import com.linkedin.camus.coders.CamusWrapper;
 import java.io.IOException;
-import org.apache.hadoop.mapreduce.RecordWriter;
+
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter;
+
+import com.linkedin.camus.coders.CamusWrapper;
 
 /**
  *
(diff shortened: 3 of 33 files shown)


bmistry13:
@HenryCaiHaiying ,

When you get chance please review or provide feedback or merge. I am more than happy to fix or changes.

Thanks,

Bhavesh

zliu41:
Hi @bmistry13 
Henry is no longer with LinkedIn. The rest of us don't have a lot of bandwidth to review Camus pull requests, so it will take longer especially for bigger requests. Thanks for contributing.

bmistry13:
Hi @zliu41 ,

Sure ! Please take your time and  here is summary of changes at high level:
1) Always close files even when one of IO Exception occurs
2) Do not append delimiter when it empty or null  for Sequence File Writer 
3) Also Failed task when partial commit camus Job (e.g job is killed via hadoop job -kill jobid_etc).
4) Most important change is if FILE name generated by the partitioner already exists ( by default it will fail)  This will ensure the data is not lost by rename  file already exists ( kafka guarantees   at-least once delivery so sticking with that default behavior) (this was root cause of all the problem of IO exceptions) 
4) For MapR Volume based HDFS directory added copy-temp-rename with retry logic.

Thanks,
Bhavesh

bmistry13:
@zliu41 ,

Do you want me breakdown down commit into smaller chunk ?  In fact, I can re-write the patch in smaller commit.  Let me know what you prefer.

Thanks,

Bhavesh

PR #201, opened 2015-04-09
write mapper start timestamp to destination folder

zliu41:
Originally the outliers are identifyed by comparing the timestamp of each avro file in the source folder, and the timestamp of the destination folder. If a source file's timestamp > destination folder's timestamp, it means this source file arrived after destination folder was created, and thus it is considered an outlier.

However, an outlier may arrive after the sweeper mappers start, and BEFORE the destination folder is created. Such an outlier will not be processed by the sweeper, and also will not be identified as outliers in the subsequent run, since its timestamp is < destination folder timestamp.

The solution is to write the sweeper MapReduce job start timestamp in the destination folder as a hidden file, e.g., /data/tracking/PageViewEvent/daily/2015/04/08/_1428543552406. This timestamp will be used to identify outliers.

### DIFF
--- a/camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
+++ b/camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
@@ -142,6 +142,10 @@ public class CamusHourlySweeper extends CamusSweeper {
     return executorService.submit(new KafkaCollectorRunner(jobName, props, errorMessages, topic));
   }
 
+  private static void createTimeStampFileInFolder(FileSystem fs, Path folder, long timeStamp) throws IOException {
+    fs.createNewFile(new Path(folder, "_" + timeStamp));
+  }
+
   private class KafkaCollectorRunner extends CamusSweeper.KafkaCollectorRunner {
 
     public KafkaCollectorRunner(String name, Properties props, List<SweeperError> errorQueue, String topic) {
@@ -192,11 +196,7 @@ public class CamusHourlySweeper extends CamusSweeper {
       CamusHourlySweeper.this.metrics.recordMrSubmitTimeByTopic(this.topicAndHour, jobSubmitTime);
       submitMrJob();
       moveTmpPathToOutputPath();
-      createTimeStampFileInOutputPath(jobSubmitTime);
-    }
-
-    private void createTimeStampFileInOutputPath(long jobSubmitTime) throws IOException {
-      this.fs.createNewFile(new Path(this.outputPath, "_" + jobSubmitTime));
+      CamusHourlySweeper.createTimeStampFileInFolder(fs, this.outputPath, jobSubmitTime);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -142,6 +142,10 @@ protected void runCollectorForTopicDir(FileSystem fs, String topic, Path topicSo
     return executorService.submit(new KafkaCollectorRunner(jobName, props, errorMessages, topic));
   }
 
+  private static void createTimeStampFileInFolder(FileSystem fs, Path folder, long timeStamp) throws IOException {
```
chavdar: Please write the create timestamp to a more generic file (Java properties or Json). We should not need to create a separate file if we want to add more metadata,

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -142,6 +142,10 @@ protected void runCollectorForTopicDir(FileSystem fs, String topic, Path topicSo
...
+    fs.createNewFile(new Path(folder, "_" + timeStamp));
+  }
+
   private class KafkaCollectorRunner extends CamusSweeper.KafkaCollectorRunner {
```
chavdar: I don't see this class in my git repo. Does this PR depend on some unmerged changes?

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -226,21 +233,57 @@ public OutlierCollectorRunner(Properties props, FileSystem fs) {
...
+   * If the destination folder already exist, get the timestamp when the MapReduce job was submitted
+   * to create that folder. The timestamp is stored in a _{timestamp} file.
+   *
+   * If such a file doesn't exist, return the timestamp of the folder.
+   * @throws IOException 
+   */
+  public static long getDestinationModTime(FileSystem fs, String outputPathStr, boolean deleteTimestamp)
+      throws IOException {
+    for (FileStatus status : fs.listStatus(new Path(outputPathStr))) {
+      if (!status.isDir() && status.getPath().getName().matches("_\\d+")) {
```
chavdar: What happens if there are multiple such files, due to failed Sweeper runs?

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Fixed. Now using a java properties file

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: It has since been renamed to CamusSingleFolderSweeper. I'll resolve conflict after review.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
zliu41: Fixed. Now when creating the timestamp file in `createStateFileInFolder()`, it first delete any existing timestamp file. So there will be at most one such file.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/CamusHourlySweeper.java
```diff
@@ -226,21 +256,69 @@ public OutlierCollectorRunner(Properties props, FileSystem fs) {
     public Void call() throws IOException {
       String inputPaths = this.props.getProperty(CamusHourlySweeper.INPUT_PATHS);
       String outputPathStr = this.props.getProperty(CamusHourlySweeper.DEST_PATH);
-      long destinationModTime = fs.getFileStatus(new Path(outputPathStr)).getModificationTime();
+      long destinationModTime = getDestinationModTime(this.fs, outputPathStr, true);
```
chavdar: why do we need to remove the file after it's been read?

zliu41: timestamp file should be removed once the outliers have been handled. Otherwise, the next run will still think those files are outliers

zliu41: Actually this is no longer needed since `createStateFileInFolder` in line 281 will first delete existing timestamp files. I've removed it.

chavdar:
LGTM

PR #203, opened 2015-04-10
Adding unit for new decoder, small change in reader to use offset from msg

ig-michaelpearce:
Hi @zliu41  these are the few small refinements on the last pull (one small change and added unit test) we had internally before code freeze COB today for production today. Have merged up from internal to share those with you.

Hope this is all good. btw I'm on annual leave as of tomorrow. So hopefully all is well with this. 

Thanks again for such a good product.


zliu41:
LGTM. thanks.

PR #205, opened 2015-04-10
Error logging before crashin

ibuenros:
Currently, Camus runs checkIfTooManySkippedMsg before reporting errors, and it crashes immediately if too many lines were skipped. Since errors in RecordReader have not been reported, it can be misterious why or which topics failed.
This pull request switches the error logging logic before the check.


zliu41:
LGTM

PR #207, opened 2015-04-13
Email notification through logging

ibuenros:


ibuenros:
@zliu41 Can you review?

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/common/EmailClient.java
```diff
@@ -0,0 +1,86 @@
...
+      props.setProperty("mail.transport.protocol", "smtp");
+      props.put("mail.smtp.host", mailHost);
+      props.put("mail.smtp.port", mailPort);
+
+      Session session = Session.getInstance(props, null);
+      MimeMessage message = new MimeMessage(session);
+      message.setFrom(new InternetAddress(senderEmail));
+
+      for (String to : emailList)
+        message.addRecipient(Message.RecipientType.TO, new InternetAddress(to, false));
```
zliu41: add braces.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
```diff
@@ -416,8 +405,10 @@ public int compare(CamusRequest r1, CamusRequest r2) {
       if (request.getEarliestOffset() > request.getOffset() || request.getOffset() > request.getLastOffset()) {
         if (request.getEarliestOffset() > request.getOffset()) {
           log.error("The earliest offset was found to be more than the current offset: " + request);
+          EmailClient.sendEmail("The earliest offset was found to be more than the current offset: " + request);
```
zliu41: Do not send email in this case. This may not be an error situation.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
```diff
@@ -416,8 +405,10 @@ public int compare(CamusRequest r1, CamusRequest r2) {
...
         } else {
           log.error("The current offset was found to be more than the latest offset: " + request);
+          EmailClient.sendEmail("The current offset was found to be more than the latest offset: " + request);
```
zliu41: Do not send email in this case.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
```diff
@@ -440,6 +431,11 @@ public int compare(CamusRequest r1, CamusRequest r2) {
                     " to start processing from earliest kafka metadata offset.");
           reportJobFailureDueToOffsetOutOfRange = true;
         }
+      } else if (3 * (request.getOffset() - request.getEarliestOffset())
+          < request.getLastOffset() - request.getOffset()) {
```
zliu41: Please send a single email for all requests, rather than one email per request.

camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/mapred/EtlInputFormat.java
ibuenros: I don't know specific use cases, so this might be wrong. Even if it is not necessarily an error situation, shouldn't we still send an email as a warning? One of these messages when a new topic is added is ok, but if they are being received often, that seems like a sign of an actual issue with the service, and the emails might be the first notice.

zliu41: if `move_to_earliest_offset=false`, then the `else` block in line 426 will be executed, the job will fail and ops will get the alert.

if `move_to_earliest_offset` is set to true, then it means whoever runs the job expects this type of situation, and shouldn't be alerted if this happens.

zliu41:
LGTM

PR #217, opened 2015-04-24
Allow s3 as possible output for destination and execution dirs

kevinsu:
allow s3 as a possible output dir for etl.destination.path, etl.execution.base.path, etl.execution.history.path.

Did this by changing the relevant FileSystem.get(conf) calls to FileSystem.get(uri, conf).  

Could use some direction on whether this is actually ok to do, and how to proceed with writing tests and such.  As it is, I'm using it to output stuff to s3 myself.  

Edit:  The reason I did this was because I was unable to get the fs.defaultFS or fs.default.name options working.  

I think a second reason, even if the fs.defaultFS stuff did work, is that this way, we are still doing the work in the local fs (hdfs for emr) and then copying data to s3, vs doing all the work in s3.  Thanks @bcotton!

### DIFF
--- a/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
+++ b/camus-etl-kafka/src/main/java/com/linkedin/camus/etl/kafka/CamusJob.java
@@ -243,12 +243,12 @@ public class CamusJob extends Configured implements Tool {
     if (getLog4jConfigure(job)) {
       DOMConfigurator.configure("log4j.xml");
     }
-    FileSystem fs = FileSystem.get(job.getConfiguration());
-
     log.info("Dir Destination set to: " + EtlMultiOutputFormat.getDestinationPath(job));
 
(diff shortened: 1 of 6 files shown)


bcotton:
FWIW, we've been using S3 and Camus for a while. We just set these in the properties file:

fs.default.name=s3://<bucket>
fs.defaultFS=s3://<bucket>

kevinsu:
I found that that did not work when using it on EMR.  I would get this error 

Exception in thread "main" java.io.IOException: Cannot initialize Cluster. Please check your configuration for mapreduce.framework.name and the correspond server addresses.

Perhaps caused by this:
2015-04-22 17:36:34,708 INFO org.apache.hadoop.mapreduce.Cluster (main): Failed to use org.apache.hadoop.mapred.YarnClientProtocolProvider due to error: Error in instantiating YarnClient

bcotton:
Could be. We are still running on Hadoop 1 in EMR.

kfleischmann:
Hey guys. i face the same Problem. How is the fix?

PR #228, opened 2015-05-13
Add expections thrown by null values to maximumDecoderExceptions.

gesundkrank:
I moved the null check up into the try block. So that exceptions caused by null values are added to maximumDecoderExceptions. Otherwise these exceptions spam the Hadoop logs.


zliu41:
LGTM. Thanks.

PR #230, opened 2015-06-05
Output more meaningful log message for camus-cleaner flows

zliu41:
Originally, the use of `topicExceptionString` was incorrect, which would often output empty exception messages.

Now it outputs those error messages as `log.error`, and the messages are rephrased to provide more information.


ydailinkedin:
LGTM.

PR #231, opened 2015-06-25
Use custom output file name that includes record count for Camus sweeper

zliu41:
Currently camus sweeper output files are named part-r-00000.avro etc
Changed to x.y.avro where x = record count in file, y = current timestamp.

### DIFF
--- a/camus-sweeper/src/main/java/com/linkedin/camus/sweeper/mapreduce/CamusSweeperOutputCommitter.java
+++ b/camus-sweeper/src/main/java/com/linkedin/camus/sweeper/mapreduce/CamusSweeperOutputCommitter.java
@@ -1,0 +1,64 @@
+package com.linkedin.camus.sweeper.mapreduce;
+
+import java.io.IOException;
+import java.lang.reflect.Method;
+
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.Counter;
+import org.apache.hadoop.mapreduce.StatusReporter;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * A subclass of org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter that uses a custom
+ * output file name: x.y.avro, where x = number of records in the file, y = current time.
+ */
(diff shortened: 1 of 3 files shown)


camus-sweeper/src/main/java/com/linkedin/camus/sweeper/mapreduce/CamusSweeperOutputCommitter.java
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }
+    super.commitTask(context);
+  }
+
+  private long getRecordCountCounter(TaskAttemptContext context) {
+    try {
+      //In Hadoop 2, TaskAttemptContext.getCounter() is available
+      Method getCounterMethod = context.getClass().getMethod("getCounter", Enum.class);
+      return ((Counter) getCounterMethod.invoke(context, AvroKeyReducer.EVENT_COUNTER.RECORD_COUNT)).getValue();
```
liyinan926: This might give you the global record count across all mappers, but what you need is the per-task record count. Have you tested this and made sure it's giving you the correct count? If not, one possible approach is to maintain the record count in the reducer and set it in the `Context` object available in the reducer class. `Reducer.Context` is a subclass of `TaskAttemptContext`. Here you can grad the value of the counter from the passed in `context`.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/utils/RelaxedAvroKeyOutputFormat.java
```diff
@@ -43,4 +50,12 @@
 
   }
 
+  @Override
+  public synchronized OutputCommitter getOutputCommitter(TaskAttemptContext context) throws IOException {
+    if (this.commiter == null) {
```
liyinan926: Why doing this?

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/mapreduce/CamusSweeperOutputCommitter.java
zliu41: The counters are maintained in Reducer class (line 42).
It has been tested in a multiple-mapper, multiple-reducer job and the result is correct.

camus-sweeper/src/main/java/com/linkedin/camus/sweeper/utils/RelaxedAvroKeyOutputFormat.java
zliu41: Lazy initialization. This is the same as the implementation in the parent method (`FileOutputFormat.getOutputCommitter`).

liyinan926:
LGTM.

PR #232, opened 2015-07-06
JSONStringMessageDecoder enhancements

szczeles:
Current implementation of JsonStringMessageDecoder allows only one field as timestamp source - if there are more than one kind of input source with different timestamping schema it does not partition messages well. In addition, field for timestamp decoding must be available in parent JSON of message.

This pull request allows to pass multiple formats of timestamp field (also nested fields) like:

    camus.message.timestamp.1.field=EventReceivedTime
    camus.message.timestamp.1.format=yyyy-MM-dd HH:mm:ss
    camus.message.timestamp.2.field=metadata.timestamp
    camus.message.timestamp.2.format=unix_milliseconds


JohannesKunzeGfK:
Hi,

when will this be merged into master?

Cheers

PR #241, opened 2015-08-13
Add parameters to control cleaner behavior

zliu41:
Added two control knobs:
- `fail.job.if.folder.not.qualify.for.deletion`: whether to fail the az job if some folders are not qualified for deletion. (default: false)
- `delete.folders.not.qualify.for.deletion`: if a folder is not qualified for deletion, whether to delete it anyway (default: true)

An hourly folder is not qualified for deletion if there are new files in the hourly folder whose timestamps are later than the corresponding daily folder. If we delete this folder there may be data loss, since these new files have not been compacted. However, if we do not delete this folder it may cause even bigger problems, since if all other hourly folders of the same day have been deleted and we don't delete this hourly folder, then in the next daily rollup, the daily sweeper will compact this folder only, and overwrite the daily folder, which may lead to bigger data loss.


zliu41:
@liyinan926 can you review?

PR #242, opened 2015-08-13
Add folder time lower bound for daily sweeper

zliu41:
Previously there was only minDaysAgo, which controls the latest allowed input folder timestamp.

Now adding maxDaysAgo. Because if camus processes an input dir that is too old, the data may be incomplete (since some of the old data may have been deleted), which may cause data loss.


zliu41:
@liyinan926 can you review?

liyinan926:
LGTM.

